PR #7, opened 2018-04-12
PAR Jérémie & Clément

Kleemdub:
Done, except from closing the connexion (when every queries are done).


ta-web-paris:
Good job, you manage to go through the iterations, looks like you understand the concept of the cruds :)
If you have questions let us know !

PR #8, opened 2018-04-12
M Rodrigo & Louis 

LouisMasurel:


ta-web-paris:
I can't review your work there is to many file (please don't submit your node module ;) )

The only thing I can say for the moment is that if you want to catch you need to define err like this : 

```js

  .catch(err => {
    console.log("Inserted failed 😡", err);
  });


```

You need to add 'err' as the parameter of the catch function, otherwise it won't catch anything :)

PR #9, opened 2018-04-12
PAR: Bertrand & Leo P.

LPsola:
@wavilo


ta-web-paris:
You are missing few iterations, if you have trouble let me know!
Otherwise good job, it look like you understand how to interact with you db :)

PR #23, opened 2018-04-12
bcn-fredrik-recipes-mongo

fredrik62:
I could only get 1 title to display, the #each method did not work as expected, not sure why, feel very lost with this.

### DIFF
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>Document</title>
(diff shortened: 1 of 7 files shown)


views/index.hbs
```diff
@@ -0,0 +1,16 @@
...
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>Document</title>
+</head>
+<body>
+ <h1>hello data will be displayed here</h1>
+  {{#each}}
```
tawebbcn: Here you should specied what you want to iterate. In your case you want to iterate the data that you are sending when you render this page whta is call it `recipe`

So it will be like 
```
{{#each recipe }}
 <p>{{this.title}}</p>
{{/each}}
```
Inside, the `this` is the placeholder for each element of the `recipe`;

PR #26, opened 2018-04-16
BCN Mihai

cosmihai:
don't know how to use findByIdAndUpdate neither findByIdAndRemove

### DIFF
--- a/models/recipe.js
+++ b/models/recipe.js
@@ -1,0 +1,45 @@
+'use strict';
+
+const mongoose = require('mongoose');
+const Schema = mongoose.Schema;
+
+const recipeSchema = new Schema({
+  title: {
+    type: String,
+    required: true,
+    unique: true
+  },
+  level: {
+    type: String,
+    enum: ['Easy Peasy', 'Amateur Chef', 'UltraPro Chef']
+  },
+  ingredients: {
+    type: Array
+  },
+  cousine: {
+    type: String,
+    required: true
+  },
+  dishType: {
+    type: String,
+    enum: ['Breakfast', 'Dish', 'Snak', 'Drink', 'Dessert', 'Other']
+  },
+  image: {
+    type: String,
+    default: 'https://images.media-allrecipes.com/images/75131.jpg'
+  },
+  duration: {
+    type: Number,
+    min: 0
+  },
+  creator: {
+    type: String
+  },
+  created: {
+    type: Date,
+    default: Date.now
+  }
+});
+
+const Recipe = mongoose.model('Recipe', recipeSchema);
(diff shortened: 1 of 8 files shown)


models/recipe.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+const mongoose = require('mongoose');
+const Schema = mongoose.Schema;
+
+const recipeSchema = new Schema({
```
tawebbcn: :+1:

recipes.js
```diff
@@ -1,11 +1,53 @@
...
-    console.log('Connected to Mongo!')
+    console.log('Connected to Mongo!');
   }).catch(err => {
-    console.error('Error connecting to mongo', err)
+    console.error('Error connecting to mongo', err);
+  });
+
+// --------adding a recipe------------
+
+Recipe.create({title: 'Scrambled eggs', level: 'Easy Peasy', ingredients: ['eggs', 'chees'], cousine: 'International', dishType: 'Breakfast', duration: 10, creator: 'Fred Flinstone'})
```
tawebbcn: :ok_hand:

recipes.js
```diff
@@ -1,11 +1,53 @@
...
+  .then((result) => {
+    console.log(result.title);
+  })
+  .catch((err) => {
+    console.log(err);
   });
 
+// ------how do I use insertMany???-----------
```
tawebbcn:
```
Recipe.insertMany(data)
  .then((data) => {
    data.forEach((recipe) => {
      console.log(recipe.title);
    });
  })
  .catch((err) => {
    console.log('An error happened:', err);
  });
```

recipes.js
```diff
@@ -1,11 +1,53 @@
...
+
+for (let i = 0; i < data.length; i++) {
+  Recipe.create(data[i], () => {
+    console.log(data[i].title);
+  });
+}
+// ---------how do I use findByIdAndUpdate---------
+// Recipe.findByIdAndUpdate('5ad51a33bc709b74dddd9ebb', {duration: 100});
```
tawebbcn: this should be working. It should be better if you do:
```Recipe.findByIdAndUpdate('5ad51a33bc709b74dddd9ebb', {$set: {duration: 100}});```

tawebbcn: I leave you here the documentation http://mongoosejs.com/docs/api.html#findbyidandupdate_findByIdAndUpdate

tawebbcn: If it's stilll not working, you can try with:

```Recipe.updateOne({ title: 'Rigatoni alla Genovese' }, { duration: 100 })```

PR #50, opened 2018-06-14
BER - Jens & Laure ThinkTank 

LaureDas:


ta-web-ber:
Did you changed the .gitignore?

Well, good job!

You just get a _undefined_ on the insertMany.
recipe is an array of objects, so you need to define which recipe :) you want to display:
```js
Recipe.insertMany(data)
	.then(recipe => {
		console.log("The user is saved and its value is: ", recipe[0].title);
	})
	.catch(err => {
		console.log("An error happened:", err);
	});
```
_Vivian_

PR #54, opened 2018-06-14
BER - Jana -done

jana-o:


ta-web-ber:
Did you changed the .gitignore?

At the end, the current disconnection doesn't work.
You should do this instead:
```js

setTimeout(() => {
	mongoose.disconnect();
	console.log("closed");
}, 5000);
```
_Vivian_

PR #55, opened 2018-06-14
BER - nesli and spela

spelavrtovec:


ta-web-ber:
Did you change the .gitignore?
I'm not able to do the correction onto your recipe.js

You better have a look on the [correction](https://github.com/ta-web-ber/the-dojo).
If there is some stuffs you don't understand, tell me.

_Vivian_

PR #121, opened 2018-07-05
MIA 07052018 Yaryna Ravlyk

YokoYolo:


recipes.js
```diff
@@ -2,10 +2,126 @@ const mongoose = require('mongoose');
 const Schema   = mongoose.Schema;
 const data = require('./data.js')
```
sandrabosk: Here you imported everything from data.js and saved it in variable data; this is what you'll use for your insertMany() method.

recipes.js
```diff
@@ -2,10 +2,126 @@ const mongoose = require('mongoose');
...
+//   })
+// .catch ((error) => {
+//   console.log (error);
+//   });
+
+
+
+
+
+  // newRecipe.insertMany([
```
sandrabosk: Reuse data variable you created on line 2.

sandrabosk:

PR #122, opened 2018-07-05
MIA christine-beck

cpbcpb:
Got this done in class.


recipes.js
```diff
@@ -1,11 +1,62 @@
...
+const recipeSchema = new Schema({
+  title: {type: String, required: true, unique: true}, 
+  level: {type: String, enum:['Easy Peasy','Amateur Chef','UltraPro Chef']},
+  ingredients: [],
+  cousine: {type: String, required: true},
+  dishType: {type: String, enum:['Breakfast','Dish','Snack','Drink','Dessert','Other']},
+  image: {type: String, default:'https://images.media-allrecipes.com/images/75131.jpg'},
+  duration: {type: Number, min:0},
+  creator: String,
+  created: {type: Date, default: Date.now}
```
```
   // fields go here
  },
  {
    timestamps:true
  }
);
```

recipes.js
```diff
@@ -1,11 +1,62 @@
...
+// })
+// .then((theRecipe)=>{
+// console.log(theRecipe)
+// })
+// .catch((theError)=>{
+
+// })
+// \
+// iteration 3
+// Recipe.insertMany(data);
```
sandrabosk: Make sure you use promises after mongoose methods (.then() nad .catch() )

sandrabosk:

PR #123, opened 2018-07-05
FTWD MIAMI Paola Arredondo - done :)

arredondop03:


recipes.js
```diff
@@ -2,10 +2,67 @@ const mongoose = require('mongoose');
...
+//   dishType: ['Dish'],
+//   image: 'https://images.media-allrecipes.com/userphotos/720x405/815964.jpg',
+//   duration: 30,
+//   creator: 'Chef Jennifer'
+// })
+
+
+
+// //insert all recipes from data.js: Iteration 3
+// Recipe.insertMany(data);
```
sandrabosk: Don't forget to use .then() and .catch().

recipes.js
```diff
@@ -2,10 +2,67 @@ const mongoose = require('mongoose');
...
+
+// //update one recipe: Iteration 4
+// Recipe.findByIdAndUpdate("5b3e627cfb68341a274268f8", {
```
sandrabosk: We really don't have id available unless we go to the DB and take one. BUt we have recipe's title available in every moment so we can use it in order to do the update:


Recipe.updateOne(
//   {title: "Rigatoni alla Genovese"}, 
//   {duration: 100}
//   )

recipes.js
```diff
@@ -2,10 +2,67 @@ const mongoose = require('mongoose');
...
+// })
+// .then((response)=>{
+//   console.log(response)
+// })
+// .catch((err)=>{
+//   console.log(err)
+// })
+
+// //remove a recipe: Iteration 5
+// Recipe.findByIdAndRemove("5b3e627cfb68341a274268f7")
```
sandrabosk: Recipe.deleteOne({title: "Carrot Cake"})

sandrabosk:
Good work and thank you for submitting.

PR #125, opened 2018-07-05
MIA Giannini Assignment completed in Collab with Will and Yara

GianniniCharles:


recipes.js
```diff
@@ -1,11 +1,67 @@
...
+const recipeSchema = new Schema({
+  title: {type: String, unique: true, required: true},
+  level: {type: String, enum: ['Easy Peasy', 'Amateur Chef', 'UltraPro Chef']},
+  ingredients: Array,
+  cousine: {type: String, required: true},
+  dishType: {type: String, enum: ['Breakfast', 'Dish', 'Snack', 'Drink', 'Dessert', 'Other']},
+  image: {type: String, default: 'https://images.media-allrecipes.com/images/75131.jpg'},
+  duration: {type: Number, min: 0},
+  creator: String,
+  created: {type: Date, default: '2018-07-05'},
```
```
   // fields go here
  },
  {
    timestamps:true
  }
);
```

sandrabosk:

PR #126, opened 2018-07-05
MIA Rodrigo Flores

rflor012:
Jackson was my partner, only missing Step 5 Remove an entry from DB


recipes.js
```diff
@@ -1,11 +1,46 @@
...
+  const recipeSchema = new Schema({
+    title: String,
+    level: {type: String, enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]},
+    ingredients: [String],
+    cousine: {type: String, required: true},
+    dishType: {type:String, enum: ["Breakfast","Dish","Snack","Drink","Dessert","Other"]},
+    image: {type: String, default:"https://images.media-allrecipes.com/images/75131.jpg"},
+    duration: {type: Number, min: [0, 'min value']},
+    creator: String,
+    created: {type: Date, default: Date.now}
```
```
   // fields go here
  },
  {
    timestamps:true
  }
);
```

sandrabosk:
Good work. Thank you.

PR #128, opened 2018-07-05
MIA Hannah Loop

hloop:
Created another pull request with Alina, but I had some errors occur with my code (and I'm not sure why).

### DIFF
--- a/recipes.js
+++ b/recipes.js
@@ -2,10 +2,89 @@ const mongoose = require('mongoose');
 const Schema   = mongoose.Schema;
 const data = require('./data.js')
 
+
+const recipeSchema = new Schema({
+  title : {type: String, required: true, unique: true},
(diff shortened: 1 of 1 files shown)


recipes.js
```diff
@@ -2,10 +2,89 @@ const mongoose = require('mongoose');
...
+      '2 green onions, thinly sliced',
+      '3 tablespoons chopped fresh cilantro leaves',
+      '2 tablespoons freshly squeezed lime juice',
+    ],
+    cousine: 'Thai',
+    dishType: 'Dish',
+    image: "https://s23209.pcdn.co/wp-content/uploads/2018/02/Easy-Thai-Red-CurryIMG_3048.jpg",
+    duration: 40 + 'mins',
+    creator: 'Kathryne',
+    created: Date('10/20/2015'),
```
```
   // fields go here
  },
  {
    timestamps:true
  }
);
```

sandrabosk:

PR #165, opened 2018-08-31
Par Aurélien & Nérimène

nerimeneakari:
@LOOFFI


recipes.js
```diff
@@ -1,11 +1,83 @@
...
+  title: {
+    type: String,
+    required: true, 
+    unique: true,
+  },
+  level: {
+    type: String, 
+    enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"],
+  },
+  ingredients: [String],
```
ta-web-paris: Array is a type as well, instead of putting the type String inside brackets, you can write it like that : 
```
ingredients: Array,
```

recipes.js
```diff
@@ -1,11 +1,83 @@
...
+  creator: String,
+  created: {
+    type: Date, 
+    default: currentDate.toDateString(),
+  }
+});
+
+const Recipe = mongoose.model("Recipe", recipeSchema);
+
+Recipe.create({title: "Poulet à la moutarde", level: "Easy Peasy", ingredients: ["Poulet", "Moutarde", "Oignons", "Sel", "Poivre"], cousine: "French", dishType: "Dish", image: "https://static.cuisineaz.com/610x610/i89759-poulet-a-la-moutarde-au-four.jpg", min: 60, creator: "Chef Nerimene" })
```
ta-web-paris: min is not a field, the field is "duration", right now the creation of the recipe works because the duration field is not required, so it creates the recipe not taking into account the "min: 60" part, but if you replace 'min' by duration' here, you will have a proper recipe in your database :)

ta-web-paris:
Good job with mongoose and queries to MongoDB !

PR #175, opened 2018-09-06
MIA FTWD Nick Crispino

crispinonicky:


MarcosPalaciosIII:
Good job on the mongoose recipe lab. The main takeaway is having knowledge of how to manipulate a database so that when you create one you can get your desired results. Also remember when your going to commit your work to have a gitignore file and add node modules to it so that your not pushing that data to github

Keep up the good work.

PR #178, opened 2018-09-06
MIA FTWD Bartashevich-Kiryl

bartosian:


MarcosPalaciosIII:
Great job on the mongoose recipe lab. The main takeaway is having knowledge of how to manipulate a database so that when you create one you can get your desired results. Also remember that it's usually a good idea to have a gitignore file and add files or folders you do not want commited to github for all to see(like node modules).

Keep up the great work.

PR #189, opened 2018-09-13
[MAD PP] Raul Martin - Christian Lopez

chrislopez24:
Fun exercise to have a first look at how we can interact with mongodb using mongoose. Looking for the next one!

### DIFF
--- a/data.js
+++ b/data.js
@@ -1,4 +1,4 @@
-module.exports =  
+module.exports =
     [
       { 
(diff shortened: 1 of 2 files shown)


ta-web-mad:
Muy buen trabajo chicos!! El mongoose.disconnect() debería estar en el último then - otherwise da un error de "topology was destroyed"

Anna

PR #257, opened 2018-11-09
Paris - Helen

hwoldu:


recipes.js
```diff
@@ -8,3 +8,69 @@ mongoose.connect('mongodb://localhost/recipeApp')
   }).catch(err => {
     console.error('Error connecting to mongo', err);
   });
+
+
+const Recipe = mongoose.model('Recipe', recipeSchema);
```
ta-web-paris: This line should come after your declare your recipeSchema, otherwise it will say that recipeSchema is not defined

recipes.js
```diff
@@ -8,3 +8,69 @@ mongoose.connect('mongodb://localhost/recipeApp')
...
+
+Recipe.insertMany(data, function(error, docs) {
+  // console.log("it worked?", docs);
+  console.log("not working", err);
+docs.forEach(pussy => {
+  console.log(`print out pussy ${pussy.title}`);
+});
```
ta-web-paris: and this block should come after you declared you Recipe variable, so if you move the other one down, you should also move this one down :)

ta-web-paris: Here, you are using the insertMany with a callback, if you want the error to be handled, you can do it like that : 
```js
Recipe.insertMany(data, function(error, docs) {
  docs.forEach(oneRecipeDoc => {
    console.log(`${oneRecipeDoc.title} added`);
  });
  if (error === true) {
    console.log("adding recipes database failed", error);
  } 
});
```
But actually is is a bit more clean to use the database query methods (like insertMAny, .crate(), find(), .findByIdAndDelete() etc.) with promises, that is to chain a .then and a .catch to check whether the query succeeded or failed : 
```js
Recipe.insertMany(data)
	.then(recipeArr => {
		recipeArr.forEach(recipe => {
			console.log(`${recipe.title} added`);
		});
	})
	.catch(err => {
		console.log("recipe.insertMany() FAILURE", err);
	});
```

ta-web-paris:
Feel free to tell me if anything is still unclear! Keep up the good work !

PR #267, opened 2018-11-13
FTWD MIA Yoli Chaviano

YoliYoli:
This was pretty easy....I think.

### DIFF
--- a/workingRecipes.js
+++ b/workingRecipes.js
@@ -1,0 +1,51 @@
+// the next three lines ara to link the files so we can use the information from them
+const mongoose = require('mongoose');
+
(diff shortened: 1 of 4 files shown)


workingRecipes.js
```diff
@@ -0,0 +1,51 @@
+// the next three lines ara to link the files so we can use the information from them
+const mongoose = require('mongoose');
+
+const Recipes = require('./Recipes');
```
ta-web-mia: const Recipe*** because this is the name that you gave to the model

ta-web-mia:
Good Job!!!!

PR #315, opened 2018-11-22
Mex - Fermin Lopez

Fermin996:
Code was being glitchy but it appears to me that I have all the necessary items. Not familiar enough to debug :/


ta-web-mex:
Good Fermín! It looks like you have missed the correct name of your routes file. That's all.
🚀 🤖 🚀 🤖 🚀 🤖

PR #367, opened 2019-01-29
[PTMIA1018] [Miller Rosas]

mrosas22:
Still missing how to close connection once the process ends automatically


ta-web-mia:
Good Job Miller!

PR #394, opened 2019-01-31
[SAO FT2 - DE - W4D3] Raphael 

RaphaelHuber:


recipes.js
```diff
@@ -1,10 +1,70 @@
...
+  });
+
+Recipe.deleteOne({ title: 'Carrot Cake' })
+  .then(() => {
+    console.log('The recipe has been deleted');
+  }).catch((err) => {
+    console.error('Error deleting recipe', err);
+  });
+
+mongoose.connection.close();
```
thebinaryfelix: Be careful with this being outside an asynchronous function like the promises above. If the operations of update, delete, insert, etc, take too long, the connection might be closed before they're done.

PR #403, opened 2019-02-01
Paris:Christina

CIBailey:


recipes.js
```diff
@@ -1,10 +1,52 @@
...
+  $push: { duration: 10 }
+})
+  .then(result => {
+    console.log("💖 Update: Success", result);
+  })
+  .catch(err => {
+    console.log("💖 Update: Failure", err);
+  });
+
+Recipe.deleteOne({ title: "Carrot Cake" })
```
ta-web-paris: If you want to delete a title you need to check to see if the title is equal to carrot cake 
`{title: { $eq: "Carrot Cake" }}`

ta-web-paris:
What you've done is good! You seem to have understood what we covered in the lesson 👏

PR #404, opened 2019-02-01
Paris SQUAD 107 - Kevin .B

kevbarns:
All queries are commented

### DIFF
--- a/models/recipe-model.js
+++ b/models/recipe-model.js
@@ -1,0 +1,34 @@
+const mongoose = require("mongoose");
+
+const Schema = mongoose.Schema;
+const recipeSchema = new Schema({
+  title: {type: String, required: true, unique: true},
+  level: {type: String, enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]},
+  ingredient: {type: []},
+  cuisine: {type: String, required: true},
+  dishType: {
+    type: String,
+    enum: ["Breakfast", "Dish", "Snack", "Drink", "Dessert", "Other"]
+  },
+  image: {
+    type: String,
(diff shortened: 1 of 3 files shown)


models/recipe-model.js
```diff
@@ -0,0 +1,34 @@
+const mongoose = require("mongoose");
+
+const Schema = mongoose.Schema;
+const recipeSchema = new Schema({
+  title: {type: String, required: true, unique: true},
+  level: {type: String, enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]},
+  ingredient: {type: []},
```
ta-web-paris: You should actually write the word array here  `type: Array`

models/recipe-model.js
```diff
@@ -0,0 +1,34 @@
...
+    type: String,
+    enum: ["Breakfast", "Dish", "Snack", "Drink", "Dessert", "Other"]
+  },
+  image: {
+    type: String,
+    default: "https://images.media-allrecipes.com/images/75131.jpg"
+  },
+  duration: {type: String, min: 0},
+  creator: {type: String},
+  created: {type: [Date]}
```
ta-web-paris: You can set a default created time by adding `default : Date.now`

ta-web-paris:
You seem to have understood the lesson 👏

When you're testing, it's important to console log both the response (res) and error (err) to see what type of data you're getting

models/recipe-model.js
kevbarns: I heard nizar saying that it should be made in the code when inserting the value, but it's great news actually ! can it be done the same way with timestamp ?

PR #407, opened 2019-02-01
Paris- Nadjie

NadjieVal:


recipes.js
```diff
@@ -1,10 +1,24 @@
...
+  level: {
+    type: String,
+    enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]
+  },
+  ingredient: [],
+  cuisine: { type: String, required: true },
+  dishType: { type: String },
+  image: {
+    type: String,
+    match: /^https?:\/\//
```
ta-web-paris: you don't need to check the match- but you can set a default here

recipes.js
```diff
@@ -1,10 +1,24 @@
...
+  },
+  duration: { type: Number, min: 0 },
+  creator: { type: String },
+  created: [Date]
```
ta-web-paris: to set the default value of when it's created to the current time you can add `default : Date.now `

PR #409, opened 2019-02-01
Paris: Poeli

pbojork:


recipes.js
```diff
@@ -1,10 +1,100 @@
...
+
+// share the "Recipe" variable with other files that require recipes.js
+// (Recipe is our Mongoose model that connects to the "recipes" collection)
+// module.exports = Recipe;
+
+// Routes
+// -------------------------------------------------------------------------------------------
+
+// (C)reate
+Recipe.create({ title: "Asian Glazed Chicken Thighs", cuisine: "Asian" })
```
ta-web-paris: You have an small error here. In your model, you say that the names need to be unique but you're using the same name as one of the items in the data. 

you get this error message "E11000 duplicate key error collection: recipeApp.recipes index: title_1 dup key: { : "Asian Glazed Chicken Thighs" }"

ta-web-paris:
What you've done is good! You seem to have understood the lesson 🙌

PR #411, opened 2019-02-01
Paris - Estelle 

Olawinski-e:


recipes.js
```diff
@@ -1,10 +1,109 @@
...
+// database. Now we need to create a Recipe Schema.
+// The schema should have the following fields:
+
+const recipeSchema = new Schema({
+  title: {
+    type: String,
+    required: true,
+    unique: true
+  },
+  level: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"],
```
ta-web-paris: for the level we want it to be an enum  

```
level: {
    type: String,
    enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]
  } 
```

an enum is like a multiple choice

recipes.js
```diff
@@ -1,10 +1,109 @@
...
+  ingredients: [String],
+  cuisine: {
+    type: String,
+    required: true
+  },
+  dishType: ["Breakfast", "Dish", "Snack", "Drink", "Dessert", "Other"],
```
ta-web-paris: dishType should also be an enum

recipes.js
```diff
@@ -1,10 +1,109 @@
...
+// array to the database. After inserting the elements,
+// print on the console the title of each recipe.
+
+Recipe.insertMany(data)
+  .then(data => {
+    console.log(data);
+  })
+  .catch(err => {
+    console.log("HOME PAGE Recipe.find() failed", err);
+    response.render("recipe-error.hbs");
```
ta-web-paris: You don't need to render anything here, because we're not running a local server or rendering pages. We didn't do any of the app.js setup

PR #412, opened 2019-02-01
Paris - Delphine

Deie:
Easy peasy 👍

### DIFF
--- a/models/recipe-model.js
+++ b/models/recipe-model.js
@@ -1,0 +1,32 @@
+const mongoose = require("mongoose");
+const Schema = mongoose.Schema;
+
+const recipeSchema = new Schema({
+  title: { type: String, required: true, unique: true },
+  level: {
+    type: String,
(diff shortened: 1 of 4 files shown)


models/recipe-model.js
```diff
@@ -0,0 +1,32 @@
+const mongoose = require("mongoose");
+const Schema = mongoose.Schema;
+
+const recipeSchema = new Schema({
+  title: { type: String, required: true, unique: true },
+  level: {
+    type: String,
+    enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]
+  },
+  ingredients: { enum: [] },
```
ta-web-paris: here, you want it to be `{ingredients: {type: Array} }` . An enum is used when there is a multiple choice

ta-web-paris:
What you've done is good! I like how you structured your code and used different files to organise your code 🎉

PR #414, opened 2019-02-01
Paris - Armand

abiglary:
Tried to add a test.js to be able to check on mongodb connection status, but it doesn't seem to work. Always return status 0 (disconnected)...


ta-web-paris:
That's great! Don't know about the test. It worths asking Nizar to know if there is a way to do a test.

PR #419, opened 2019-02-02
PARIS - Bastien "done"

basttran:
Is ugly but is working. Can't wait for the solution :)

### DIFF
--- a/recipes.js
+++ b/recipes.js
@@ -1,10 +1,94 @@
-const mongoose = require('mongoose');
-const Schema   = mongoose.Schema;
-const data = require('./data.js');
+const mongoose = require("mongoose");
+const Schema = mongoose.Schema;
+const data = require("./data.js");
+const Recipe = require("./models/recipe-model.js");
 
-mongoose.connect('mongodb://localhost/recipeApp')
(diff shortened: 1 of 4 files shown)


recipes.js
```diff
@@ -1,10 +1,94 @@
...
+function batchInsert() {
+  Recipe.insertMany(data)
+    .then(recipeDoc => {
+      console.log("Recipe BATCH INSERT success!", recipeDoc.__title);
+      batchUpdate();
+    })
+    .catch(err => {
+      console.log("Recipe BATCH INSERT failure!", err);
+    });
+}
```
ta-web-paris: You did the exercise in a more complex way. But that's totally ok! That was what we were waiting for (very simple)

```
Recipe.insertMany(data)
  .then(recipeDoc => {
    console.log("All the recipes from data.js have been pushed 🥗", recipeDoc);
  })
  .catch(err => {
    console.log("Push the data.js file Failed! 🍎🍐", err);
  });
```

ta-web-paris:
That's great! You wrapped everything in functions to reuse them that is really handy! :) That's not a mess at all! You did exactly what you should do! 👍

PR #420, opened 2019-02-03
Paris - Samuel

petitkriket:
I believe I forgot to make a pull request and i can't find a clue about pulls history..
Just in case,  it's done and here (minus a premise on iteration 6)

### DIFF
--- a/recipes.js
+++ b/recipes.js
@@ -1,10 +1,110 @@
-const mongoose = require('mongoose');
-const Schema   = mongoose.Schema;
-const data = require('./data.js');
+const mongoose = require("mongoose");
+const express = require("express");
+const app = express();
+app.use(require('easy-livereload')());
 
-mongoose.connect('mongodb://localhost/recipeApp')
-  .then(() => {
-    console.log('Connected to Mongo!');
-  }).catch(err => {
-    console.error('Error connecting to mongo', err);
-  });
+//const Schema = mongoose.Schema;
+const chalk = require("chalk");
+
+const data = require("./data.js");
+const Recipe = require("./models/recipe-model.js");
+
+mongoose
+	.connect("mongodb://localhost/recipeApp")
+	.then(() => {
+		console.log(chalk.green("Connected to Mongo! ❤️"));
+	})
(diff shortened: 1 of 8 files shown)


recipes.js
```diff
@@ -1,10 +1,110 @@
...
+const Recipe = require("./models/recipe-model.js");
+
+mongoose
+	.connect("mongodb://localhost/recipeApp")
+	.then(() => {
+		console.log(chalk.green("Connected to Mongo! ❤️"));
+	})
+	.catch(err => {
+		console.log(chalk.red.bold("Error connecting to Mongo....... 💩", err));
+	});
```
ta-web-paris: good idea to use chalk!

recipes.js
```diff
@@ -1,10 +1,110 @@
...
+
+// set server
+app.listen(3000, () => {
+	console.log("expressive llamas in pyjamas baby");
+});
```
ta-web-paris: love your console log! 
The app.listen is not mandatory because the exercise was just manipulating mongoose. But it's no big deal though. :)

recipes.js
```diff
@@ -1,10 +1,110 @@
...
+		if (err) {
+			console.log("Something wrong when updating data!");
+		}
+
+		console.log(chalk.magenta.bold("Recipe successfully updated to:"));
+		console.log(chalk.white.dim(doc));
+	}
+);
+
+Recipe.deleteOne({
```
ta-web-paris: You could use the `findByIdAndUpdate()` function for this in case there is two recipes with the same name it would delete both. Same for the update function.

ta-web-paris:
Very good! I like the fact you added everything in a website. That makes things easier to correct even if it was optionnal! 👍

recipes.js
petitkriket: aha I love [easter eggs](https://dev.to/philnash/developer-whimsy-llamas-in-pajamas-232i) !

petitkriket:
I prefer to build actual things even if it's only training and dummy stufff ! It's like makings legos, satisfying !

PR #433, opened 2019-02-12
Cameron Chromia FTWD MIA

cameron-chromiak:
iz done

### DIFF
--- a/recipes.js
+++ b/recipes.js
@@ -21,7 +21,7 @@ const recipeSchema = new Schema({
   date: {type: Date, default: Date.now}
 });
 
-const Recipe  = mongoose.model('Recipe', recipeSchema);
+const Recipe  = mongoose.model('recipe', recipeSchema);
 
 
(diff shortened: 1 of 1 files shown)


ta-web-mia:
Great job so far! Your create and insertMany methods are all working perfectly, but the findByIdAndUpdate isn't doing anything at the moment. Try changing line 49 of your recipes.js to the following:

`  Recipe.updateOne({title: "Rigatoni alla Genovese"}, {duration: 100})
`

It's important to realize that an object's ID in Compass is unique, and a different one is created with each object. That means that even if two objects are identical, they will always have different IDs. By extension, you can only ever change one object at a time when using the findByIdAndUpdate method. 

You can also add the following line of code to the bottom of recipes.js to disconnect as soon as you've updated your database:

` setTimeout(() => { mongoose.connection.close(); console.log("connection closed"); }, 1000);`

So far so good, keep up the good work 👍

PR #436, opened 2019-02-12
"Monroe Getz FTWD Miami"

Mgetz10:


ta-web-mia:
Well done! All of the methods are reflected properly in the Compass database, and the only thing that can really be improved on is closing the connection once everything is finished. If you try to use

` mongoose.connection.close();`

on its own, it will end up closing the connection before any of the methods. To circumvent this, you need to use setTimeout to delay the disconnection. Replace lines 86 through 93 on the recipes.js file with the following code:

` setTimeout(() => { mongoose.connection.close(); console.log("connection closed"); }, 1000);`

The methods are by far the most important takeaway from this lab, and you managed to work through every single one. Keep up the good work 👍

PR #437, opened 2019-02-12
Liran Gibor FTWD-Miami 2019

LiranGib:


ta-web-mia:
Well done! All of the methods are reflected properly in the Compass database, and the only thing that can really be improved on is closing the connection once everything is finished. If you try to use

`    mongoose.connection.close();`

on its own, it will end up closing the connection before any of the methods. To circumvent this, you need to use setTimeout to delay the disconnection. Add the following code to the end of your recipes.js and it should work properly:

`  setTimeout(() => {
    mongoose.connection.close();
    console.log("connection closed");
  }, 1000);`

The methods are by far the most important takeaway from this lab, and you managed to work through every single one. Keep up the good work 👍

PR #438, opened 2019-02-12
ric&henry&nunna.TFWD miami

nunna19:


ta-web-mia:
Great job, guys! The only issue I saw was that the recipe you created to add to compass didn't end up being added. Since every later iteration works perfectly, so I'd assume that you had a working version of it at some point earlier. Nothing to worry about! Keep up the good work 👍

PR #654, opened 2019-08-28
PAR FTWD HenriDeBel

belke05:


ta-web-paris:
Good job! Just make sure to put valid values according to your model - when you create a recipe, "Lunch" is not a dishType in your model

PR #658, opened 2019-08-28
PARFTWD Aimée & Sophie-Anne

sophieannebourgine:


ta-web-paris:
Recipe.findByIdAndUpdate("Rigatoni alla Genovese", { duration: "100" })
  .then(recipe => {
    console.log("duration:", recipe);
  })
  .catch(err => {
    console.log("An error happened:", err);
  });

This will not work, findById needs an Id, either you use 
Recipe.updateOne({title: Rigatoni alla Genovese"}, { duration: "100" })
  .then(recipe => {
    console.log("duration:", recipe);
  })
  .catch(err => {
    console.log("An error happened:", err);
  });

either you take its id in the database and use it in the function

And becareful when closing the database, the deleteOne function didn't complete and the database has been closed before

PR #741, opened 2019-09-12
MIA - FTWD Osheit Chang

osheitchang:


ta-web-mia:
Keep your Mongoosing strong, it's one of the best skills to have as a full stack developer. But for what you did, good job!

PR #780, opened 2019-11-06
WDFT Paris Venkat

venkat-raghulan:


ta-web-paris:
Great job! 
You could also create one function per action and call them in then statement. It would be more readable and you could reused it after

PR #781, opened 2019-11-06
[PAR WDFT] Hanna

hanna974:


index.js
```diff
@@ -1,12 +1,58 @@
...
+  creator: "Jo"
+})
+  .then(recipe => {
+    console.log(`Recipe title is: ", ${recipe.title}`);
+  })
+  .catch(err => {
+    console.log("An error happened:", err);
+  });
+
+Recipe.update({ title: "Rigatoni alla Genovese" }, { $set: { duration: 100 } })
```
ta-web-paris: you don't need the $set as the second argument of update is the value to update

index.js
```diff
@@ -1,12 +1,58 @@
...
+
+Recipe.deleteOne({ title: "Carrot Cake" })
+  .then(recipe => {
+    console.log(`Suppression done: ${recipe.title}`);
+  })
+  .catch(err => {
+    console.log("An error happened:", err);
+  });
+
+mongoose.connection.close();
```
ta-web-paris: be careful to comment your function otherwise each time you saved it will run them

ta-web-paris:
Good job!

PR #783, opened 2019-11-06
[PAR-FTWD]-LOUISE

LouiseMarissal:
done with the frank s help
Thank's

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,50 @@
-const mongoose = require('mongoose');
-const Recipe = require('./models/Recipe'); // Import of the model Recipe from './models/Recipe'
-const data = require('./data.js');  // Import of the data from './data.js'
-
+const mongoose = require("mongoose");
+const Recipe = require("./models/Recipe"); // Import of the model Recipe from './models/Recipe'
+const data = require("./data.js"); // Import of the data from './data.js'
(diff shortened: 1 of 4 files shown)


models/Recipe.js
```diff
@@ -1,9 +1,37 @@
-const mongoose = require('mongoose');
+const mongoose = require("mongoose");
 const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
-  // TODO: write the schema
+  // TODO: write the schema -----
+  // Iteration 1
+  title: String,
```
ta-web-paris: it is better to make it unique

index.js
```diff
@@ -1,12 +1,50 @@
...
-    console.log('Connected to Mongo!');
-  }).catch(err => {
-    console.error('Error connecting to mongo', err);
+    console.log("Connected to Mongo!");
+  })
+  .catch(err => {
+    console.error("Error connecting to mongo", err);
   });
+// iteration 2
+Recipe.create({
```
ta-web-paris: you function are good but you should call them inside promise.
When you launch mongoose.connect you can directly start calling function in the then so you embed promises.

ta-web-paris:
Good job anyway!

PR #784, opened 2019-11-06
[PAR-FTWD]Camille & Helene

Millcat:
all done with .then and .catch promises method + asyn way of franck - titles of each recipes console.log as well

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,60 @@
 const mongoose = require('mongoose');
 const Recipe = require('./models/Recipe'); // Import of the model Recipe from './models/Recipe'
-const data = require('./data.js');  // Import of the data from './data.js'
+const data = require('./data.js'); // Import of the data from './data.js'
 
 // Connection to the database "recipeApp"
-mongoose.connect('mongodb://localhost/recipeApp', { useNewUrlParser: true })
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,12 +1,78 @@
...
+  })
+  .then((createdRecipe) => {
+    console.log(createdRecipe);
+    console.log(createdRecipe.title);
+    return Recipe.insertMany(data);
+
+  }).then(recipes => {
+    recipes.forEach(recipe => console.log(recipe.title));
+
+  }).then(() => {
```
ta-web-paris: here you can directly call findandupdate in the then of insertMany

models/Recipe.js
```diff
@@ -2,8 +2,45 @@ const mongoose = require('mongoose');
...
+  image: {
+    type: String,
+    default: 'https://images.media-allrecipes.com/images/75131.jpg'
+  },
+  duration: {
+    type: Number,
+    min: 0
+  },
+  creator: {
+    type: String
```
ta-web-paris: you can also write: 
creator: String

ta-web-paris:
Great job girlz !

PR #801, opened 2019-11-10
[PAR-FTWD] Véro & Marie

Mawee33:
iterations 1 to 5 (iteration 6 with Franck during renforcement


index.js
```diff
@@ -10,3 +10,31 @@ mongoose.connect('mongodb://localhost/recipeApp', { useNewUrlParser: true })
     console.error('Error connecting to mongo', err);
```
ta-web-paris: you can also need all your CRUD function inside the mongo connect and when they are all done you can disconnect your DB.

ta-web-paris:
Good Job!

PR #844, opened 2019-11-17
LIS-FTWD-Dennis-174

Denn-is:
@ironhackLis-webTeam

had lots of problems with this, coudl you please address:
1. is the general structure of this correct?
- nesting everything into the successful connection and putting the catch for the mongoDb connection at the end
- repeating Recipe every iteratio (should it just be then?)
2. order of execution: how can i tell it to execute iteration 1 first and iteration 5 last?
3. why does mongoose conenction close produce an error in the utils?
4. why does deleting everything at the beginnen produce a success AND an error?
5. why is document.title undefined?
6. how can i list every documnt (the entire result-database) at the end?

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,9 @@ mongoose
   })
   .then(() => {
     console.log("Connected to Mongo!");
+    Recipe.deleteMany({})
+      .then(console.log("SUCCESS: database has been cleared"))
+      .catch(console.log("ERROR: database could not be cleared"));
     // Create 1 recipe(hardcode):
     Recipe.create({
       title: "Pie14",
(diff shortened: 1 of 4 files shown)


Denn-is:
and worst off all: the compass still doesn't connect!

ironhackLis-webTeam:
Hi dennis.

To get the compass running try this:

1. open the activity monitor (Aktivitätsanzeige) on your Mac.
2. go to CPU in the tap. and search for any active mongo db activity.
3. Kill the process  (big button with x in top left corner)

4. then go to your command line and run this command:
mongod --config /usr/local/etc/mongod.conf

5. Open the compass and press connect.

Let me know if that works. And otherwise we can have another look at it
together.

In the meantime i'll have a look at your code for this lab.

Paula :)

ironhackLis-webTeam:
Regarding your code:

1. You are not required to nest every Iteration with your Recipe model inside the mongoose connect promise. Try to keep in mind that the code is red from top to bottom so this will effect the way in which your code gets executed and should also answer your 2. question as well. 

3. /4.  review the syntax that you are using after the .deleteMany. this is why it throws you both of the error messages

``Recipe.deleteMany({})
      .then(console.log("SUCCESS: database has been cleared"))
      .catch(console.log("ERROR: database could not be cleared"));``

Try this instead: 

``.then(() => {
        console.log("SUCCESS: database has been cleared");
      })
      .catch(() => {
        console.log("ERROR: database could not be cleared");
      });
});``

this is the major part that was breaking your code and is also fixing the document.title (5.) from being undefined.

6. to list through every document (the lab is asking for the titles only) you can use a map function like this: 

``Recipe.insertMany(data)
  .then(recipe => {
    recipe.map(item => {
      console.log(item.title);
    });
  })
  .catch(error => {
    console.log(error);
  });``


hope this helps, let me know if you have more questions. :)
Paula

PR #860, opened 2019-12-12
AMS - 201912 - MONIKA SWIDZINSKA

mswidzinska:
6 iterations (with close the database)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,62 @@
 const mongoose = require('mongoose');
 const Recipe = require('./models/Recipe'); // Import of the model Recipe from './models/Recipe'
-const data = require('./data.js');  // Import of the data from './data.js'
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,12 +1,62 @@
...
+        console.log("Carrot Cake deleted");
+    })
+    .catch((err) => {
+        console.log("Error", err);
+    })
+    .then(() => mongoose.connection.close())
+    .then(() => console.log("----------DATABASE CLOSED-------------"))
+    .catch(err => {
+        console.error('Error connecting to mongo', err);
+    });
```
elyasp: Separating the promises from their return values is one way of doing. However to improve the modularity of our code we could use Promise.all in this case. Pushing every promise in an array passed as an argument to Promise.all("arg") will do the same thing as above and our code will be easier to understand. ;)

PR #862, opened 2019-12-12
[AMS WDFT 112019] - Michael Ta

miketaart:


index.js
```diff
@@ -1,12 +1,92 @@
...
+  title: 'King Prawn Noodles',
+  level: 'UltraPro Chef',
+  ingredients: ['Noodles', 'Prawns', 'Soya', 'Oil'],
+  cuisine: 'Asian',
+  dishType: 'Dish',
+  duration: 30,
+  creator: 'Michael Ta', 
+})
+
+.then((KingPrawnNoodles)=> {
```
elyasp: unused redundant argument

index.js
```diff
@@ -1,12 +1,92 @@
...
+  })
+  .catch((error)=> {
+    console.log("Err", error)
+  })
+
+  // https://github.com/Piepongwong/async-javascript
+
+//Iteration 6
+mongoose.connection.close(() => {
+  console.log("Disconnected")
```
elyasp: closing mongoose connection like "mongoose.connection.close()" return a promise.
Therefore this won't work as you need to catch its success and error. Appending it with .then will work.

PR #895, opened 2020-01-29
[PAR-WD-FT] - Nina ongoing

Gorlanova:


clarabesset:
Great job on this lab!

One small detail: when updating a recipe, it is better to find it (here, "Rigatoni alla Genovese") by title rather than by ObjectId since the latter is generated by mongo automatically and will change each time you change computer or insert the database. Since you put "unique: true" in the model for the title field, you made sure that targetting "Rigatoni alla Genovese" is as precise as using an id to do so :)

Gorlanova:
Ah nice ! Thank you !!

PR #896, opened 2020-01-29
[PAR-WDFT-0120] Yannick Bourenane - ongoing

yannick-bourenane:


Jepassaispar:
All good :)
Even if in the lab they ask you to use `DeleteOne` method, I would always advise you to use `findOneAndDelete` because this is the method that will get in the catch whenever it won't find the data to delete.
It is very good that you put your calls inside functions and that you are closing the db each time, but then you would have to connect each time which can be tedious, in a project it will be always better to either use the await method or nest your promises in other promises if you want to fight the asynchronicity :)
But that's small picks, your lab is great !

yannick-bourenane:
Hey thanks for the feedback, i'll probably try to rewrite this (if i find
some time) with the async function to practice it :)

PR #897, opened 2020-01-29
[PAR-WDFT-0120] Pierre Etryhard ongoing

PierreEtryhard:


clarabesset:
Great job on this lab!

One small thing: index.js, line 45 you tried to console.log "title" but is not accessible since "data" is the array of recipes, you'd need to use a forEach loop for example, to access each title inside the array

also, when updating a recipe, it is better to find it (here, "Rigatoni alla Genovese") by title rather than by ObjectId since the latter is generated by mongo automatically and will change each time you change computer or insert the database. Since you put "unique: true" in the model for the title field, you made sure that targetting "Rigatoni alla Genovese" is as precise as using an id to do so :) - same for delete

PR #920, opened 2020-01-30
[PAR-WDFT-0120] - Mehmet

MehmetCan75:


clarabesset:
Good job on this lab!

You'd just need to console.log the title of each recipe at the same time (in the then) for the insertMany for it to be perfect, so we know which one we inserted at which point.

PR #925, opened 2020-01-30
[BER-FT-012020] Jacqueline C.

chenjacqueline:


index.js
```diff
@@ -1,13 +1,83 @@
...
+    console.log("Recipe has been updated successfully!");
+  }
+);
+
+Recipe.deleteOne({ title: "carrot cake" }).then(() => {
+  console.log("Recipe has been deleted!");
+});
+
+setTimeout(() => {
+  mongoose.connection.close();
```
fenrew: It's fine to do it like this, another way to do it is to put it at the end of the last mongoose "promise" (see above)

index.js
```diff
@@ -1,13 +1,83 @@
...
+  });
+
+Recipe.updateOne({ title: "rigatoni alla genovese" }, { duration: 100 }).then(
+  () => {
+    console.log("Recipe has been updated successfully!");
+  }
+);
+
+Recipe.deleteOne({ title: "carrot cake" }).then(() => {
+  console.log("Recipe has been deleted!");
```
fenrew: mongoose.connection.close()

PR #929, opened 2020-01-30
[BER-FT] Elsa

ElsaFil:
I put everything inside the completion of the first "promise", to make sure all the recipes are already created before we edit/delete them. Is that the correct way to do it? 
Thanks! 💫


fenrew:
So the solution is that is has to be chained like you did in class, otherwise it is not sure that the mongoose.insertMany will be completed before we start doing the updateOne and delete. So you did it correctly though it's better to chain the .then()  :)

ElsaFil:
thanks a lot!

PR #935, opened 2020-01-30
[BER FT] Tim

timbotimber:
Done it all!

Really went super fast today. Would be great to recap quickly the over-arching concepts tomorrow morning in the lecture ( like what express does etc) just to re-enforce things.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,37 @@
-const mongoose = require('mongoose');
-const Recipe = require('./models/Recipe.model'); // Import of the model Recipe from './models/Recipe.model.js'
-const data = require('./data.js'); // Import of the data from './data.js'
+const mongoose = require("mongoose");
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,13 +1,38 @@
...
+
+Recipe.updateOne({ title: "Rigatoni alla Genovese" }, { duration: 100 })
+  .then(result => {
+    console.log("success!");
+  })
+  .catch(err => {
+    console.log(err);
+  });
+
+Recipe.deleteOne({ title: "Carrot Cake" }).then(result => {
```
DalinaLeila: your catch is missing here

timbotimber:
> looks good! 👍 If you want to recap, its always good to go back to the learning platform and read through the things after you learned them in class. It helped me a lot :)

good idea! Will do that at the weekend

PR #939, opened 2020-01-30
[BER-FT-012020]) - Saeka 

saekaisozumi:
I have a question.

In Iteration 3, "print on the console the title of each recipe."
How to access to data.js?
I thought //console.log(data.title); works, but it doesn't.
Do I need to make for loop...?

Except that, I hope everything is okay:)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,46 @@
-const mongoose = require('mongoose');
-const Recipe = require('./models/Recipe.model'); // Import of the model Recipe from './models/Recipe.model.js'
-const data = require('./data.js'); // Import of the data from './data.js'
+const mongoose = require("mongoose");
+const Recipe = require("./models/Recipe.model"); // Import of the model Recipe from './models/Recipe.model.js'
+//console.log(Recipe);
+const data = require("./data.js"); // Import of the data from './data.js'
 
 // Connection to the database "recipeApp"
 mongoose
-  .connect('mongodb://localhost/recipe-app-dev', {
+  .connect("mongodb://localhost/recipe-app-dev", {
     useCreateIndex: true,
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
-  .then(x => console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`))
-  .catch(err => console.error('Error connecting to mongo', err));
+  .then(x =>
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,13 +1,46 @@
...
+  cuisine: "French",
+  dishType: "Breakfast",
+  image: "https://images.media-allrecipes.com/images/75131.jpg",
+  duration: 3,
+  creator: "John Doe",
+  created: new Date()
+}); */
+
+//Recipe.insertMany(data);
+//console.log(Recipe.title);
```
fonsogms: here you should have console.log after the then().
example:
```
Recipe.insertMany(data)
  .then(createdRecipe => {
    console.log(createdRecipe);
  })
  .catch(err => {
    console.log(err);
  });
```
Something like that would have worked

index.js
```diff
@@ -1,13 +1,46 @@
...
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
-  .then(x => console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`))
-  .catch(err => console.error('Error connecting to mongo', err));
+  .then(x =>
+    console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`)
+  )
+  .catch(err => console.error("Error connecting to mongo", err));
+
```
fonsogms: Instead of commenting the create method you could have drop the collection at the beginning everytime you run your code.
like this `Recipe.collection.drop()`
this way you wouldnt get the duplicate error cause you would clean the database at the beginning every time

index.js
```diff
@@ -1,13 +1,46 @@
...
+  result => {
+    console.log("Success!");
+  }
+);
+
+Recipe.deleteOne({ title: "Carrot Cake" }).then(result => {
+  console.log("Success!2");
+});
+
+setTimeout(() => {
```
fonsogms: this is a nice workaround with the close connection problem of the dastabase but it is better practice using    mongoose.connection.close() inside the then function of one of the last methods you use, such as deleteoone
example:
```
Recipe.deleteOne({ title: "Carrot Cake" }).then(result => {
  console.log("Success!2");
   mongoose.connection.close();
});
```
This will close the database after you delete whatever you wanted to delete

PR #940, opened 2020-01-30
[BER, FTWD, 012020], Mohamed Ciwan

ciwan-aram:


index.js
```diff
@@ -2,12 +2,52 @@ const mongoose = require('mongoose');
...
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
-  .then(x => console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`))
+  .then(x =>
+    console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`)
+  )
   .catch(err => console.error('Error connecting to mongo', err));
+
+// // Recipe.create({
```
fonsogms: instead of commenting to avoid overlapping your insertions you could have tried before the create method clean the collection everytime your run your code.
Something like this:
```
Recipe.collection.drop()
```
This way you will clean your collection at the beginning of everytime your run the code

ciwan-aram: yes! I learned it, thank you for feedback :)

PR #941, opened 2020-01-30
[BER-WD-FT-012020]-Victoria

VictoriaBlat:
Markus the Pizza Creator

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,43 @@
-const mongoose = require('mongoose');
-const Recipe = require('./models/Recipe.model'); // Import of the model Recipe from './models/Recipe.model.js'
-const data = require('./data.js'); // Import of the data from './data.js'
+const mongoose = require("mongoose");
+const Recipe = require("./models/Recipe.model"); // Import of the model Recipe from './models/Recipe.model.js'
+const data = require("./data.js"); // Import of the data from './data.js'
 
 // Connection to the database "recipeApp"
 mongoose
-  .connect('mongodb://localhost/recipe-app-dev', {
+  .connect("mongodb://localhost/recipe-app-dev", {
     useCreateIndex: true,
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
-  .then(x => console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`))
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,13 +1,43 @@
...
+  result => {
+    console.log(result);
+  }
+);
+
+Recipe.deleteOne({ title: "Carrot Cake" }).then(deleted => {
+  console.log(deleted);
+});
+
+mongoose.connection.close();
```
fonsogms: carfeul with closing the connection in the global scope.
Since all of this functions are asyncronous the close() method can happen before the other methods such as create, delete or update.
This will create a lot of problems becuase your connection  could close before any of the other methods are done so it will give you an error such as this
```
UnhandledPromiseRejectionWarning: MongoError: server is closed
    at executeWriteOperation (/private/tmp/lab-mongoose-recipes/node_modules/mongodb/lib/core/sdam/server.js:470:14)
```

index.js
```diff
@@ -1,13 +1,43 @@
...
+
+Recipe.updateOne({ title: "Rigatoni alla Genovese" }, { duration: 100 }).then(
+  result => {
+    console.log(result);
+  }
+);
+
+Recipe.deleteOne({ title: "Carrot Cake" }).then(deleted => {
+  console.log(deleted);
+});
```
fonsogms: Also I encourage you to use the catch method as well to be able to see the erros if your database crashes.
It will save you a lot of trouble in the future :)
Errors are our friends

index.js
```diff
@@ -1,13 +1,43 @@
...
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
-  .then(x => console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`))
-  .catch(err => console.error('Error connecting to mongo', err));
+  .then(x =>
+    console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`)
+  )
+  .catch(err => console.error("Error connecting to mongo", err));
+
```
fonsogms: If you dont want your data to overlap in the database everytime you run the code you can drop the collection at the beginning like this:
```
Recipes.collection.drop()
```

PR #943, opened 2020-01-31
[BER-WDFT-012020] - Shahinya

sha-hinya:


index.js
```diff
@@ -3,11 +3,68 @@ const Recipe = require('./models/Recipe.model'); // Import of the model Recipe f
...
+  duration: 40,
+  creator: 'Chef Sha'
+}).then(recipe => {
+  console.log(recipe.title);
+}).catch(err => {
+  console.log(err);
+});
+
+
+Recipe.collection.drop();
```
fonsogms: is there any reason why you didnt put the drop method at the beginning of the database management?
This way you would have started with a clean database at the begining of your database.

PR #945, opened 2020-01-31
[PAR][WDFT][Lab mongoose recipes][TAYSSIR]done ongoing

taytay81:


clarabesset:
Good job on this lab! 

Just watch line 28, you wrote "foreach" instead of "forEach" (native js functions are case sensitive), small mistake though, nice work!

PR #946, opened 2020-01-31
[BER-FT] Laurel Wright

lwrocks:
Done!


index.js
```diff
@@ -1,13 +1,66 @@
...
+    }).catch(err => {
+      console.log(err);
+    });
+
+  })
+  .catch(err => {
+    console.log(err);
+  });
+
+  mongoose.connection.close();
```
fonsogms: If you used  `mongoose.connection.close();` in the global scope you run the risk of closing your database before updating it, because these are asyncronous methods which will happen depending if the promise has been resolved or not.
So you will have this type of errors:
`MongoError: server is closed`
A good solution could be adding the    mongoose.connection.close(); inside of your last method, which is the deleteOne :)

PR #948, opened 2020-02-01
[PAR-WDFT] Louis 01/02/2020

louis-md:


index.js
```diff
@@ -1,13 +1,41 @@
...
+
+const deleteRecipe = Recipe.deleteOne({title: 'Carrot Cake'})
+  .then (deletedRecipe => console.log(`${deletedRecipe.title} has been deleted!`))
+  .catch(error => console.log(`Ooh! An error happened while deleting the recipe: ${error}`))
+
+Promise.all([createNewRecipe, importRecipes, editRecipe, deleteRecipe])
+  .then(values => {
+    console.log(`Life is magic! All promises were filled: ${values}`);
+    mongoose.connection.close();})
+  .catch(error => console.error(`Sadness, something went wrong resolving all promises: ${error}`))
```
ta-web-paris: All promises are asynchronous, so when your try to delete and edit your recipes, they might not have been created yet, so at the end they are not being updated.

Promise.all runs your promises simultaneously, not in the respective order they are in the array.

You could run Promise.all with the createNewRecipe and the importRecipes since they do not depend on each other, but you'd have to put the editRecipe / deleteRecipe only once the other 2 functions have been run, and then close the DB.
Good catch on the Promise.all though

PR #949, opened 2020-02-01
PAR WEBDEV 0120 MORSLI PAULINE

PaulineLalou:


index.js
```diff
@@ -11,3 +12,25 @@ mongoose
...
+
+  // Recipe.insertMany(data, function(error, docs) {});
+
+  // Recipe.updateOne({ title: 'Rigatoni alla Genovese' }, { duration: 100 })
+  // .then()
+  // .catch();
+  
+  Recipe.deleteOne({ title: 'Carrot Cake' })
+  .then(console.log("success, you deleted Carrot Cake !"))
+  .catch();
```
ta-web-paris: I noticed you don't log your errors in the catch block, it would be better if you did as you'd know where / which promise failed along the line.
It doesn't matter for the purpose of this lab but on a bigger project with many lines of code it would be easier to track down errors.

PR #950, opened 2020-02-03
PAR WDFT0120 Anthony L ongoing

anth0sama:


clarabesset:
Watch out, in your model you forgot "const mongoose = require('mongoose'); const Schema = mongoose.Schema;", when you reference them in the rest of the file it doesn't know what you're talking about and you'll get an "undefined". 

Also there are some quick fixes to make on line 25 in index.js :) I'll leave this open, when you get time, make the changes and comment here !

PR #953, opened 2020-02-04
[PAR WDFT] - Shannon, done ongoing

tranquildoe:


clarabesset:
Good work! It's great you kept the notes from Franck to look back on it later on !

PR #982, opened 2020-02-13
Short refactors, add exempt pull request labels

josecarneiro:
Made several clarifications on the exercise explanation.

Changed the format of the data file from `js` to `json`. It exposes the students to the fact that they can require the contents of a JSON file.

Added exempt PR labels to stalebot config.


josecarneiro:
Added step to clean the database before performing any operations, to avoid issues with uniquely indexed properties.

PR #1007, opened 2020-03-10
Anastasiia Kozlova PT Amsterdam

anastezia-kz:


index.js
```diff
@@ -1,13 +1,67 @@
...
+  ],
+  cuisine: "Asian",
+  dishType: "Dish",
+  image: "https://images.media-allrecipes.com/userphotos/720x405/815964.jpg",
+  duration: 40,
+  creator: "Chef LePapu"
+};
+Recipe.create(recipe)
+  .then(recipe => console.log(`I took the recipe from data, ${recipe.title}`))
+
```
LloydC: You should have a .catch(error => console.log(error)) here as well ;)

PR #1008, opened 2020-03-10
final commit

chelseybee19:


index.js
```diff
@@ -11,3 +22,28 @@ mongoose
   })
   .then(x => console.log(`Connected to Mongo! Database name: "${x.connections[0].name}"`))
   .catch(err => console.error('Error connecting to mongo', err));
+
+  Recipe.create(recipe)
+  .then(recipe => console.log(` ${recipe.title}`));
+
```
LloydC: You should have a .catch(error => console.log(error)) here just in case ;)

PR #1079, opened 2020-04-15
PAR WDFT Sam L ongoing

samlfair:


TatijanaRajcic:
Hey Sam! Thanks for your lab! I like the fact that you used async/await. Small things:

- There a small error while I reach the manyRecipes.forEach : i have the error "cannot read property forEach of undefined": it does not like your "manyRecipes" ! Why so? because in insertManyRecipes, you should only return the promise, without the empty .then.

Your function becomes:

```
function insertManyRecipes() {
  return Recipe.insertMany(manyRecipes)
}
```

Then, when you use a variable (here, manyRecipes) you actually store the result of this Promise if it succeeds, so in order words **you don't need extra .then() to catch the result of a promise if you are using async/await**. 

Otherwise it's perfect! Great job!

samlfair:
Thank you!

TatijanaRajcic:
To be more precise: you did not need an extra .then in your case, because you were working on the result of your promises in your recipesProcesses function. But you could have written the whole thing differently: see this example I just came accross:

```
// Connection to the database "recipe-app"
mongoose
  .connect(MONGODB_URI, {
    useCreateIndex: true,
    useNewUrlParser: true,
    useUnifiedTopology: true
  })
  .then(self => {
    console.log(`Connected to the database: "${self.connection.name}"`);
  })
  .then(() => {
    steps()
  })

  .catch(error => {
  });



  // TRYING ASYNC 

const newRecipe = ()=>{
  return Recipe.create(myRecipe).then((myRecipe) => {
    console.log(`The recipe was saved ${myRecipe.title}`)
  })
}

const insertManyData = () =>{
  return Recipe.insertMany(data).then((recipes) => {
    recipes.forEach((recipe) => {
      console.log(`${recipe.title}`)
    })
  })
}

const updateData = ()=>{
  return Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese"},{ duration: 100},{new: true}).then((dbresp) => {
    console.log(`Duration updated on recipe ${dbresp.title}`)
  })
}

const deleteData = () =>{
  return Recipe.deleteOne({title: 'Carrot Cake'}).then(recipe =>{
    console.log(`Recipe deleted!`);
  })
}

async function steps () {
  try{
    await newRecipe (); 
    await insertManyData (); 
    await updateData(); 
    await deleteData();
    mongoose.connection.close(() => {
      console.log("mongoose is now closed");
    }) 
  } 
  catch(err){
    console.log(err)
  }
}
```

Hope it makes sense!

PR #1092, opened 2020-04-15
PAR-FT-032020 Yuval Bugajski ongoing

YuvalBu:


TatijanaRajcic:
Hey Yuval! Nice lab! Since I know you attended the reinforcement, I think now you understand well how to imbricate .then()

Keep in mind that right now you are console.loging every time the full result of your request to the database (i.e. when you create a recipe, what you get as a result from the DB is an object of the newly created recipe, when you insert many recipes what you get as a result from the DB is an array of the newly created recipes, etc). Every method from Mongoose returns something specific (you can check the doc to see what they return, or console.log the result as you do, to see what you get), but nothing stops you to "rework" on this DBResult. 

For exemple, in the .then of the "create", what you get in the console.log is a object of the newly created recipe. Instead of console.loging dbres, you could simply do console.log(dbres.title) if you wanted to retrieve only the title.

Nice work anyway! If you have time, give async/await a try !

PR #1094, opened 2020-04-15
PAR WDFT 202003 - Anne Patana - Week4 Day3 ongoing

apatana:


TatijanaRajcic:
Hey Anne! Nice job! It's almost perfect!
- Be careful to indent properly your code, so that we don't get lost. You are missing a tab from line 19
- Your code regarding the database close connection does not get executed. I guess the last piece of code comes from the doc, but I imagine the use case was different, because right now it does not work. What you could do instead is closing the database, right after you are sure you did all the operations you want i.e. after deleting the Carrot Cake recipe. Therefore, this part of your code (mongoose.connection.close()) can go inside the .then of the Delete operation.

```
          Recipe.deleteOne({title: 'Carrot Cake'})
          .then((dbResponse) => {
            console.log("Carrot Cake has been removed from the database");
            mongoose.connection.close(() => {
              console.log('Mongoose default connection disconnected through app termination');
            });
          })
```

PR #1105, opened 2020-04-17
[PARIS-FT-042020] - Benjamin Dean

bdean05:
My MR


ta-web-paris:
Hey Benjamin !
Nice job on this lab,

Just ont thing : be careful in your model : 
- the instructions said "title" must be unique and required, that means you need to specify it in your model (with keys like required: true / unique: true), same for "cuisine" :)

Otherwise good !

Nina

PR #1120, opened 2020-04-19
[PAR WDFT] Nelly BP

LuxiaAnon:
Good morning teachers 😪 

Not a big fan of promises for now 😅 

Wish you a very good day ! 🌻

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,4 @@
 const mongoose = require('mongoose');
-
 // Import of the model Recipe from './models/Recipe.model.js'
 const Recipe = require('./models/Recipe.model');
 // Import of the data from './data.json'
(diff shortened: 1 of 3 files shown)


ta-web-paris:
Hello Nelly ! 
No worries about promises, you'll get them with practice ! 
For basics then / catch promises remember the two ways :

- Chained promises : a lot of then, and one catch
>> Model.action 1 (blabla)
.then 1 (blabla + action 2)
.then 2 (blabla + action 3)
.then 3 (blabla + action 4)
.catch X (catches one error for all, not very precise)

- Nested promises : as many then / catch as you have actions : 
>> Model.action 1 (blabla)
>> .then 1 (blabla +
>>>> action 2 (blabla)
>>>> .then 2 (blabla + 
>>>>>> action 3 (blabla)
>>>>>> .then 3 (blabla)
>>>>>> .catch 3 (blabla)

>>>> .catch 2 (blabla)

>>.catch 1 (blabla)

In your case, you used the nested method. You should place the mongoose.close inside the then of delete (which is inside the then of update (which is inside the then of insertMany)))
That means one action waits for another to be executed before happening (you close the database after deleting one item, and you delete one item after updating another, and you update after importing recipes, etc etc...)

it's pretty messy but there's a clean way for all this : async await ! it allows you to tell the action to (A)WAIT for something before executing (ASYNChronously) !

Hope it helps... Don't worry you'll get it soon ! Back end is a lot of information you just have to digest everything haha

Hit me up for questions
And thank you for your lovely notes in pull requests :) :) :)

Nina

PR #1268, opened 2020-07-30
WEB - REM - 2007 Alvaro, Lanette, Bodei

BodeiBrouwer:


index.js
```diff
@@ -20,8 +20,52 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    Recipe.create({title: "Asian Glazed Chicken Thighs",
```
IH-WebDev-TA-Remote, line 23: Just for good practices, remember to indent the objects and put all properties new lines (If they are multiline objects). Just for having a more clear code. You can also create the object before the promise inside a variable and then just pass that variable inside the promise.

index.js
```diff
@@ -20,8 +20,52 @@ mongoose
...
+    creator: "Chef LePapu"})
+      .then((result) => {
+        console.log(result.title)
+      })
+      .catch ((err) => {
+        console.log('Loser!', err)
+      })
+
+    Recipe.insertMany(data)
+    .then((thing) => {
```
IH-WebDev-TA-Remote, line 48: using thing is fine when you have no idea what you are getting xD but after you know it is the same recipes you created you can change this parameters to be descriptive of the content. Something like recipes or insertedRecipes. You may be using this variables later in some additional coding and for readability it is much better.

index.js
```diff
@@ -20,8 +20,52 @@ mongoose
...
+      thing.forEach((elem) => {
+        console.log(elem.title)
+      })
+      Recipe.deleteOne({title: 'Carrot Cake'})
```
IH-WebDev-TA-Remote, line 52: watch indentation here.

IH-WebDev-TA-Remote:
All promises are good and well structured, however they are all being executed at the same time, and because they are promises, they might resolve at completely different times and order. Meaning perhaps the promise to delete a recipe might resolve before the promise to add all recipes. And because of that, you might encounter a problem.
There are several ways to go around this so you will have to try the one you feel more confortable with.
1. Putting the next promises inside the .then of the previous one.
2. Using promiseAll to resolve the promises you need to resolve first, and inside the .then of promiseAll, creating the ones you want to resolve after.
3. chaining several .then inside the mongoose connection (this might not work in future labs/projects as the mongoose connection will be stablished in a separate file)
For the closing of the mongoose connection is it just using the promise ( mongoose.connection.close() ), however, because it is a promise itself, you will need to make sure all of the previous promises resolved first using any of the methods explained in order to close the mongoose connection only after everything else was resolved.
To finish, the lab solution is good, syntax is fine, the schema was perfect but it is important that you understand more how the promises work between then. Keep up the good work!

PR #1271, opened 2020-07-30
[WDFT-REM-2007] Joel and Elise

elisedjn:
That was challenging! The last iteration broke our brains but it seems to work 😅

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,35 @@ mongoose
   })
   .then(() => {
     // Run your code here, after you have insured that the connection was made
+    // Iteration 2
+    Recipe.create(data[0])
+      .then(() => {
+        Recipe.find({}, {title: 1})
+          .then((result) => console.log(result));
+      })
+    
+    // Iteration 3
+    Recipe.insertMany(data)
+      .then(() => {
+        Recipe.find({}, {title: 1})
+          .then((result) => {
+            console.log(result)
+            
+            // Iteration 4
+            Recipe.findOneAndUpdate({title: "Rigatoni alla Genovese"}, {$set:{duration: 100}})
+              .then(() => {
+                console.log('Value Updated')
+
+                // Iteration 5
+                Recipe.deleteOne({title:"Carrot Cake"})
+                  .then(() => {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -21,7 +21,35 @@ mongoose
   })
   .then(() => {
     // Run your code here, after you have insured that the connection was made
+    // Iteration 2
+    Recipe.create(data[0])
+      .then(() => {
+        Recipe.find({}, {title: 1})
```
IH-WebDev-TA-Remote, line 27: Using the .find here is good for being extra sure it is in the DB but might actually not be needed. The .create promise will give you back the element it created inside the parameter of its .then(). So you can you console.log it from there.

index.js
```diff
@@ -21,7 +21,35 @@ mongoose
...
+          .then((result) => console.log(result));
+      })
+    
+    // Iteration 3
+    Recipe.insertMany(data)
+      .then(() => {
+        Recipe.find({}, {title: 1})
```
IH-WebDev-TA-Remote, line 34: Same as before, the .find might be redundant here.

index.js
```diff
@@ -21,7 +21,35 @@ mongoose
...
+            // Iteration 4
+            Recipe.findOneAndUpdate({title: "Rigatoni alla Genovese"}, {$set:{duration: 100}})
+              .then(() => {
+                console.log('Value Updated')
+
+                // Iteration 5
+                Recipe.deleteOne({title:"Carrot Cake"})
+                  .then(() => {
+                    console.log("Recipe deleted")
+                    mongoose.connection.close()
```
IH-WebDev-TA-Remote, line 47: .close is a promise itself, so as all promises, they should be with a then and a catch if the connection was closes correctly or if there was an error.

IH-WebDev-TA-Remote:
Very good job! very clever nesting the promises to manage the order they resolved!
I would say the only important feedback is to add the .catch to your promises, this is not only good practice for checking errors in your server terminal but sometimes you will use this information to send errors to the user.
But again, overall, very very good job! keep it up :)

PR #1272, opened 2020-07-30
[WEB-REM-2007] Rafael and Gonzalo

rlaz133:


index.js
```diff
@@ -20,8 +21,25 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    Recipe.create({title: 'Caesar Salad', level: 'Easy Peasy', cuisine: 'Mediterranean'})
+      .then(()=>{
+        Recipe.find({title: 'Caesar Salad'}, {title:1})
```
IH-WebDev-TA-Remote, line 26: the create method will return the created element, so you can use a parameter on it's .then() and access the information from there. There is no need to run a .find()

index.js
```diff
@@ -20,8 +21,25 @@ mongoose
...
+          .then((recipe)=>
+          console.log(recipe)
+          )
+      })
+    Recipe.insertMany(data)
+      .then(()=>{
+        Recipe.find({},{title:1})
```
IH-WebDev-TA-Remote, line 33: Same as previous comment

PR #1273, opened 2020-07-30
[WDFT-REM-2007] Marta & Maeva

maevamerrou:


index.js
```diff
@@ -1,27 +1,74 @@
...
     // Before adding any documents to the database, let's delete all previous entries
-    return self.connection.dropDatabase();
+    return self.connection.dropDatabase(); // auto delete everything to avoid doublons
   })
+
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+
+    Recipe.create({
+      title: "Tortilla",
```
IH-WebDev-TA-Remote, line 29: You could save the object in a variable before the promise and the just insert the variable here. Usually helps with readability :)

index.js
```diff
@@ -1,27 +1,74 @@
...
+      level: "Easy Peasy",
+      ingredients: "patatoes, eggs, oliv oil, onion, salt, peper",
+      cuisine: "Spanish",
+      dishType: "main_course",
+      image: "https://www.cocinaabuenashoras.com/files/tortilla-de-patatas-espanola-768x576.jpg",
+      duration: 25,
+      creator: "Manish",
+      created: "",
+    })
```
IH-WebDev-TA-Remote, line 38: .create is a promise, so it should had it's .then and .catch . Even as per instructions, you can have a parameter inside the then, and console.log the title of the object created.

index.js
```diff
@@ -1,27 +1,74 @@
...
+
+    Recipe.insertMany(data) 
+    .then(recipe => { recipe.map(item => { console.log(item.title); }); })
```
IH-WebDev-TA-Remote, line 41: if you are only doing a console.log you can use a .forEach() instead of a .map(). Remember .map() does the same as a forEach() but also returns an array with all the elements. In this case there is no need for the array and you are not even saving it inside a variable. So using a simple forEach would be better in this case.

index.js
```diff
@@ -1,27 +1,74 @@
...
+      // Recipe.insertMany(data)
+    
+    // })
+      // .then((res) => {
+      //   data.forEach((recipe)=> console.log(recipe.title));
+      // })
+
+  
+      Recipe.findOneAndUpdate({title: 'Rigatoni alla Genovese'}, { $set: {duration: 100}})
+        .then(success => {
```
IH-WebDev-TA-Remote, line 53: the parameter inside the .then() of findOneAndUpdate and deleteOne methods is not necesary a success, these will vary depending on the method. For one it can be the element removed, for the other the element updated. Remember to be more descriptive of your variables and parameters. And in this case you are not using the variables so you might as well not put them there.

IH-WebDev-TA-Remote:
Good job, the syntax is correct for all promises. However one big issue you might encounter is that all the promises (including the mongoose.connection.close() ) are all in the same place. Meaning they will be called at the same time but might resolve in a different order of the one they are being called. To solve this there are severals approaches.
1. Putting the promises you want resolved after in the nested .then() of the the previous promises.
2. Saving promises in variables and using promiseAll to wait until they all resolve until the new ones are called.
3. Chaining .then() of the mongoose connection, so the promises inside those separate .then() will resolve in order.
This is very important when working with promises so I would recommend you to try again using any of the above methods and you can even go through the solution we will send later.
Keep up the good work! :)

PR #1276, opened 2020-07-31
lab2

GunnerAg:
(Alejandro & Gunner).
Could not find what was wrong with the last 2 on the documentation.


index.js
```diff
@@ -1,27 +1,66 @@
...
+  })
+  .then(self => {
+    console.log(`Connected to the database: "${self.connection.name}"`);
+    // Before adding any documents to the database, let's delete all previous entries
+    // return self.connection.dropDatabase();
+  })
+  .then(() => {
+    // Run your code here, after you have insured that the connection was made
+
+    Recipe.create({
```
IH-WebDev-TA-Remote, line 25: You could save the object inside a variable before the promise and then just add the variable inside the promise. Helps with readability. Also before you are using the first element of the data.json you could just target that via bracket notation. like data[0]

IH-WebDev-TA-Remote:
Schemas are perfect and easy to read. There are some issues with proper indentation in index.js. Promises are in the same .then() but with different indentations.
Overall the syntax is good, just a couple of important feedback. 
First remember to add .then() and .catch() to your promises. This is important, specially in the development process to make sure everything is working correctly and if errors might be coming from the database.
Second, all your promises are in the same level, meaning they will execute at the same time, but, because they are promises, they might resolve in different order. So if the delete one resolved before you add the elements, you will definitely get an error. To solve this you need to use any of the following 3 steps:
1. Nest your promise calls inside the .then() of the previous ones.
2. Save the promise in variables and use promiseAll to wait for the previous ones to resolve before the next ones.
3. Chain several .then() of the mongoose connection to put the different promises, this will ensure the all run in the correct order. However this solution might not be the best moving forward as you will see the mongoose connection will be made separately.
It is very important that you understand how promises resolve at different times and the reason managing this is essential. Consider going back to this lab during the weekend and trying one of the proposed solutions or just going through the solution.
Keep up the good work!

PR #1279, opened 2020-07-31
REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:


index.js
```diff
@@ -19,8 +19,15 @@ mongoose
     // Before adding any documents to the database, let's delete all previous entries
     return self.connection.dropDatabase();
   })
-  .then(() => {
+  .then(async (self) => {
```
itstheandre, line 22: here, the self is uneceesary :)
the if you are not using it no need to "declare it"

PR #1285, opened 2020-07-31
[REMOTE WDPT 0602] Lynn W-B

lynn-w-b:
All iterations completed but...
Iteration 6 is not working although I don't know why? The code looks right to me but every time I run it I get lots of lines of error - throw err: Cannot read property 'map' of undefined.???

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
 mongoose
   .connect(MONGODB_URI, {
     useCreateIndex: true,
+    useFindAndModify: false,
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
@@ -37,11 +38,11 @@ mongoose
       .catch(error => console.log(`Error occurred deleting recipe: ${error}`));
     })
 
-    .then(() => {
+    /*.then(() => {
       mongoose.connection.close()
       .then(() => console.log('Connection with database closed successfully'))
       .catch(error => console.log(`Error disconnecting from Mongoose:${error}`));
-        })
+        })*/
       
   .catch(error => {
     console.error('Error connecting to the database', error);
--- a/index.js
+++ b/index.js
@@ -20,8 +20,29 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -20,8 +21,29 @@ mongoose
...
-  })
+    Recipe.create({title:'Cold Chocolate Drink', level:'Easy Peasy', ingredients: ['oat milk','raw cacao'], cuisine:'Raw Foods', dishType: 'drink', duration: 2, creator: 'LWB'})
+      .then(recipe => console.log(`The recipe was saved and its title is: ${recipe.title}`))
+      .catch(error => console.log(`An error happened: ${error}`));
+    
+    Recipe.insertMany(data)
+      .then (recipes => recipes.forEach(recipe => console.log(`Title: ${recipe.title}`)))
+      .catch (error => console.log(`Error occurred adding recipes to database:${error}`));
+      
+    Recipe.findOneAndUpdate({title:'Rigatoni alla Genovese'},{$set:{duration:100}}, {new:true})
```
valmgisbert, line 32: Your findOneAndUpdate already knows that the second key is the one to update once it recognises it, so your $set is unnecessary here ($set usually adds new fields to a document; it can also update, but it is used more with the method .aggregate())

index.js
```diff
@@ -20,8 +21,29 @@ mongoose
...
+      .then (success => console.log(`Recipe successfully updated`))
+      .catch( error => console.log(`Error occurred updating recipe: ${error}`));
+    
+    Recipe.deleteOne({title:'Carrot Cake'})
+      .then (deleted => console.log(`Recipe successfully deleted`))
+      .catch(error => console.log(`Error occurred deleting recipe: ${error}`));
+    })
+
+    /*.then(() => {
```
valmgisbert, line 41: In here, you're treating this section as a chained promise, and since you're not returning anything your .then() is rendered useless in this case; it would work if either you chain your .then()s/make promise.all() and start returning your methods inside your .then()s, or if you insert the command to close the connection in the final CRUD method, OR if you just call it as another block of code after your CRUD methods. All of these options can be seen in the solutions posted in the class sheet. If you need any extra explanation, let me know!

PR #1286, opened 2020-07-31
completed lab

AndrewMastercode:


index.js
```diff
@@ -25,3 +25,45 @@ mongoose
...
+    console.log(`Creating a new recipe went wrong! Try again 😞 ${err}`)
+  );
+
+Recipe.insertMany(data)
+.then(recipeDocs => console.log(`new recipes inserted: ${recipeDocs}`))
+.catch(err =>
+  console.log(`Inserting new recipes went wrong! Try again 😞 ${err}`)
+);
+
+Recipe.updateOne({title:"Rigatoni alla Genovese"},{$set: {duration:100}})
```
valmgisbert, line 59: Your findOneAndUpdate already knows that the second key is the one to update once it recognises it, so your $set is unnecessary here ($set usually adds new fields to a document; it can also update, but it is used more with the method .aggregate())

index.js
```diff
@@ -25,3 +25,45 @@ mongoose
...
+    ],
+    cuisine: "Asian",
+    dishType: "main_course",
+    image: "https://images.media-allrecipes.com/userphotos/720x405/815964.jpg",
+    duration: 40,
+    creator: "Chef LePapu"
+  })
+  .then(recipeDoc => console.log(`The title is: ${recipeDoc.title} and this is the whole recipe:${recipeDoc}`))
+  .catch(err =>
+    console.log(`Creating a new recipe went wrong! Try again 😞 ${err}`)
```
valmgisbert, line 50: Great in sticking with template literals throughout the file!

valmgisbert:
Great job, Andre! really good structure; if you're feeling adventurous you could refactor it into a chain of promises to practice with the promise.all(). Your last iteration, the closing of the connection, can be put inside of the deleteOne to maintain continuity in your CRUD cycle, or you can do it after this method.

PR #1287, opened 2020-07-31
[REMOTE-PT-0620] Debora Crescenzo

NubivagantYansa:
- I added iterations 2-5 in a async() to make sure promises are resolved in order (I wouldn't want to delete a recipe before it's added to the database). I'd curious to know an alternative way to solve that.

- I left only one for.Each/ console.log commented to show I understood how to iterate through the array of recipes instances (plus the default that were already in the exercise). Before deleting the console log I checked that everything was working as expected.

- I found the 'disconnect' method on the documentation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,27 +1,60 @@
-const mongoose = require('mongoose');
+const mongoose = require("mongoose");
 
 // Import of the model Recipe from './models/Recipe.model.js'
-const Recipe = require('./models/Recipe.model');
+const Recipe = require("./models/Recipe.model");
 // Import of the data from './data.json'
-const data = require('./data');
+const data = require("./data");
 
-const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,27 +1,60 @@
...
+      //   )
+      // )
+      .catch((err) =>
+        console.log(`Failed adding receipes from data. Err: ${err}`)
+      );
+
+    await Recipe.findOneAndUpdate(
+      //Iteration 4: update value duration to 100
+      { title: "Rigatoni alla Genovese" },
+      { $set: { duration: 100 } },
```
valmgisbert, line 46: Your findOneAndUpdate already knows that the second key is the one to update once it recognises it, so your $set is unnecessary here ($set usually adds new fields to a document; it can also update, but it is used more with the method .aggregate())

valmgisbert:
Wonderful work, Debora! your use of async await is amazing here. We're gonna share some solutions in a while that use different methods so you can compare and practice.

PR #1289, opened 2020-08-01
[REMOTE-WDPT-0602] -Cristina Matamoro

cris-developer:
Done everything.

1. Although i was connected to the DB, and i am console logging succesfully the updating and deletions on the terminal, they are seen on Compass. Why?.  Can you help me out? The recipes, and the new one created i can see it updated on compass.

2. While closing mongoose (iteration 6), i retrieved the below error. How can i fix it?

/Users/cris/Desktop/IRONHACKHOMEWORK/lab-mongoose-recipes/node_modules/mongodb/lib/utils.js:106
      throw err;
      ^
TypeError: Cannot read property 'map' of undefined

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,27 @@ const data = require('./data');
 
 const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
 
+const myRecipe = {
+  title: "Galician Soup Broth",
+  level: 'Easy Peasy',
+  ingredients: [
+    "8 ounce (225 grams) dry white beans",
+    "32 ounce (1 liter) water",
+    "1 ham bone",
+    "1 veal (or beef) bone",
+    "1 ounces (20 gram) pork fat",
+    "2 handfuls fresh turnip leaves",
+    "2 pounds (1 kilogram) potatoes",
+    "salt to taste"
+  ],
+  dishType:"soup",
+  cuisine: "Galician",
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -19,9 +40,40 @@ mongoose
...
-  })
+     Recipe
+      .insertMany(data)
+      .then(totallistrecipes => totallistrecipes.forEach(recipe => console.log(recipe.title)))
+      .catch(error => console.error('Inserting multiple recipes error:', error))
+    }) 
+  //Iteration 4 - Update recipe
+  .then(() => {
+    Recipe
+    .findOneAndUpdate({title: 'Rigatoni alla Genovese'}, { $set: {duration: '100'}}, {new: true})
```
valmgisbert, line 61: Your findOneAndUpdate already knows that the second key is the one to update once it recognises it, so your $set is unnecessary here ($set usually adds new fields to a document; it can also update, but it is used more with the method .aggregate())

valmgisbert:
What I see in your code is that you're trying to set is as chained promises, but you're still writing is as blocked sections of your CRUD methods (if your .then()s don't return anything if they're being chained, they're rendered useless). Also, you should practice template literals in your console.logs/returned strings. In the shared solutions you have both approaches that you were trying to achieve. If you have any questions, let me know!

PR #1309, opened 2020-09-03
WEBDEV-BERLIN-KevinKlein

pkevinklein:


pkevinklein:
for some reason I am not able to connect with the db. Therefore, I cannot check if the changes are being accordingly executed. Help!

PR #1347, opened 2020-09-17
[PAR - WDFT - 248] Timothée NICOLE

timothee-nicole:


Fanny-Sbn:
Hello Timothée, thanks for submitting your lab. 

Recipe model 👍 

Nice use of async await function. Just a few comment, you need to disconnect the db in your lauchCRUD function and run your launchCRUD function at line 23 because you need to be sure you're connected to the db first and then make your operations and close your db. 

Good job 😄

PR #1360, opened 2020-09-17
[PAR-WDFT-248]Laura Caillaux

Lcdevi:
I have to admit that all these are not very clear for me right now... :/

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,9 +19,39 @@ mongoose
     // Before adding any documents to the database, let's delete all previous entries
     return self.connection.dropDatabase();
   })
+  // .then(() => {
+  //   // Recipe.create(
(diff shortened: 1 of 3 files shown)


Fanny-Sbn:
Hello Laura, thanks for your lab.

In your Recipe Model, you have twice image, remove line 14.

Actually your code is working well, the only mistake is for the console log of the recipes titles. You console logged directly the recipes titles from data, but you need to it with the data you got from the database so, it should look like this : 

Recipe.insertMany(data)
    .then(recipes => console.log(recipes.map(recipe=>recipe.title)))

You also could have used async await function, have a look to the correction when it will be in dev-squad repo 😉 

Good job Laura !

PR #1361, opened 2020-09-17
[PAR-WDFT-248] Ailie McCorkindale

RaspberryChaos:


ta-web-paris:
Hey Ailie ! 
Nice job on this lab, just one thing : you don't have to do as many async as you did, you can do just one async and several await inside :) 

Otherwise all good

Nina

PR #1362, opened 2020-09-18
[PAR WDFT] Laurent Vuillaume

figura01:
done

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,15 +12,81 @@ mongoose
   .connect(MONGODB_URI, {
     useCreateIndex: true,
     useNewUrlParser: true,
-    useUnifiedTopology: true
-  })
-  .then(self => {
+    useUnifiedTopology: true,
+    useFindAndModify: true
+  }).then(self => {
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hello Laurent !

Nice use of testing, but two things are correct in your lab :
- you have to create a recipe with all the values (yours is just a "test" title it's a little light)
- you have to define more parameters in your Schema : entries have a type but they also have other attributes (are they required ? are they unique ? etc) so you need to work on that too

You should try and correct these points before I mark this lab as fully completed !
You can as well make the tests more compact (maybe just one at the end with a list of the recipes titles ? to check if those who are supposed to be here are here indeed ? it should be enough)

You can try to work with async await as well instead of "then" if you want (though it's not mandatory)

Nina

figura01:
Ty Nina for your comments, in fact I don't have read enough the lab, I miss
the part with the rules of model, ('' name is unique.. ") so I will rework
on that 😋, ty

PR #1363, opened 2020-09-18
[PAR-WDFT-248] SALIM AMEUR

ameursalim:


ta-web-paris:
Hey Salim ! 
Nice job here ! Just one small thing, in your update function, you don't specify which recipe is supposed to be updates (usually you can find it by its name since it's unique it's like an ID, and then set the duration time)

Otherwise nice !

Nina

PR #1368, opened 2020-10-08
[REM-2009] Sandra + Sumi

scmendez:
We tried our best to make pretty, but any feedback on how to improve would be greatly appreciated! (Although we know you're gonna be super busy with lab #2 feedback)

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,14 +14,71 @@ mongoose
     useNewUrlParser: true,
     useUnifiedTopology: true
   })
+
   .then(self => {
     console.log(`Connected to the database: "${self.connection.name}"`);
     // Before adding any documents to the database, let's delete all previous entries
     return self.connection.dropDatabase();
   })
+
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    Recipe.create({
+        title: 'recipe1',
+        cuisine: "Asian"
+      })
+      .then((response) => {
+        console.log(response)
+      })
+      .catch((err) => {
+        console.log('oh no:', err)
+      })
+
+    Recipe.insertMany(data)
+      .then((response) => {
+        response.forEach(recipe => {
+          console.log(recipe.title)
+        })
+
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -14,14 +14,71 @@ mongoose
...
+    Recipe.insertMany(data)
+      .then((response) => {
+        response.forEach(recipe => {
+          console.log(recipe.title)
+        })
+
+        let updateRigatoni = Recipe.findOneAndUpdate({
+          title: 'Rigatoni alla Genovese'
+        }, {
+          $set: {
```
IH-WebDev-TA-Remote, line 45: findOneAndUpdate knows you will be updating a property in the second argument. So you don't need to specify this with the $set.

index.js
```diff
@@ -14,14 +14,71 @@ mongoose
...
+            duration: 100
+          }
+        })
+
+        updateRigatoni.then(() => {
+          console.log('rigatoni updated!')
```
IH-WebDev-TA-Remote, line 51: you can put the code in these .then and .catch inside the promiseAll .then and .catch. And they act the same way. So checking the resolution of the promises here and not in the promiseAll is sort of redundant.

index.js
```diff
@@ -14,14 +14,71 @@ mongoose
...
+
+        removeCarrotCake.then(() => {
+          console.log('carrot cake removed!')
+        }).catch((err) => {
+          console.log('carrot cake not removed', err)
+        })
+
+        Promise.all([updateRigatoni, removeCarrotCake])
+          .then(() => {
+            console.log('connection is closed')
```
IH-WebDev-TA-Remote, line 68: the console.log will happend before the mongoose connection is closed. 
mongoose.connection.close() is itselft a promise so the console.log to confirm this happened should be inside a .then attached to the connection.close() promise.

IH-WebDev-TA-Remote:
Very good job! Check individual feedback for some improvement, mostly on some redundant code. But still, overall everything works correctly. Well done :)

PR #1370, opened 2020-10-08
[REM 2009 ] Jordane, Katie and Ross

kmstamey:


index.js
```diff
@@ -16,12 +16,52 @@ mongoose
...
+      //createRecipe
+      .then((response) => {
+        console.log(response)
+      })
+    let recipeMany = Recipe.insertMany(data)
+    recipeMany
+      .then(() => {
+        console.log('all added')
+      })
+    Promise.all([recipeMany]) //iteration 4
```
IH-WebDev-TA-Remote, line 33: promise.all is mostly used when you want several promises to resolve. In this case, since you only need one to resolve you can put the code inside this .then inside the promise .then on line 30 instead.

index.js
```diff
@@ -16,12 +16,52 @@ mongoose
...
+      .then(() => {
+        let updatePromise = Recipe.updateOne({ title: "Rigatoni alla Genovese" }, { $set: { duration: 100 } })
```
IH-WebDev-TA-Remote, line 35: for mongoose method that update files like this one (updateOne) you don't need to use $set, as the method will always expect what you want to update as the second argument.

index.js
```diff
@@ -16,12 +16,52 @@ mongoose
...
+        updatePromise
+          .then(() => {
```
IH-WebDev-TA-Remote, line 37: the .then of this promise might be redundant because you already have a .then on the promise.all below. You can simply save the promise into a variable and then run the console.log inside the promise.all .then.

index.js
```diff
@@ -16,12 +16,52 @@ mongoose
...
+            console.log('Updated worked!')
+          })
+        let deletePromise = Recipe.deleteOne({ title: 'Carrot Cake' })
+        deletePromise
+          .then((response) => {
```
IH-WebDev-TA-Remote, line 42: same as above, see the solution we will send later for how to approach this.

index.js
```diff
@@ -16,12 +16,52 @@ mongoose
...
+
+          })
+          .catch((err) => {
+            console.log('Not deleted')
+            console.log(err)
+          })
+        Promise.all([updatePromise, deletePromise])
+          .then(() => {
+            mongoose.connection.close()
+              .then((response) => {
```
IH-WebDev-TA-Remote, line 53: if you are not using the response inside this .then, then there is no need to put it as a parameter.

IH-WebDev-TA-Remote:
Good job, and I see you made good use of promise.all to close the connection (although with a little code redundancy in there) Check the specific feedback sessions and the solutio we will send later so how to better work around promise.all.
Still, overall everything works good. Well done!

PR #1371, opened 2020-10-08
[REM 2009] Vykintas and Lovro

LovroHodak:


index.js
```diff
@@ -21,7 +23,55 @@ mongoose
...
   .then(() => {
     // Run your code here, after you have insured that the connection was made
+   /* THIS IS ITERATION 2!!!!!!!!!
+    Recipe.create( data[0] )
+      .then((recipe) => {
+        console.log(recipe.title)
+
+      })
+      .catch(()=> {
+        console.log()
```
IH-WebDev-TA-Remote, line 33: don't console.log the nothingness. Instead catch and error and console.log that :p

index.js
```diff
@@ -21,7 +23,55 @@ mongoose
...
+      })
+      */
+
+     Recipe.insertMany(data)
+        .then((recipes) => {
+          //console.log(recipes)
+          for(let recipe in recipes){
```
IH-WebDev-TA-Remote, line 40: good use of for in loop, but instead of "recipe" a better variable name would be index.

index.js
```diff
@@ -21,7 +23,55 @@ mongoose
...
+            console.log(recipes[recipe].title)
+          }
+
+          let updatedPromise = Recipe.updateOne({title: "Rigatoni alla Genovese"}, {$set: {duration: 100}})
```
IH-WebDev-TA-Remote, line 44: for mongoose methods that update the data. you don't need to use $set, as it already knows the second argument that you pass will be the data to be updated.

index.js
```diff
@@ -21,7 +23,55 @@ mongoose
...
+
+            updatedPromise.then(() => {
+                console.log('Updated yo!')
```
IH-WebDev-TA-Remote, line 47: since you are already using promise.all later you don't need to check if the promise resolved here as well. Instead you can put this console.log inside the .then of promise.all

index.js
```diff
@@ -21,7 +23,55 @@ mongoose
...
+            })
+            .catch((err) => {
+              console.log(err)
+            })
+            let deletedPromise = Recipe.deleteOne({title: 'Carrot Cake'})
+
+              deletedPromise.then(() => {
+                console.log('Deleted yo!')
```
IH-WebDev-TA-Remote, line 55: same as before, this part of the code is a little redundant since you are using promise.all later for the same promise.

IH-WebDev-TA-Remote:
Very well done, good use of promise.all, however there is some code redundancy in there. Check the solution we will send later to see how best to approach this.
Still overall, good job. Keep it up!

PR #1372, opened 2020-10-08
[REM 2009] Mohammed Imran && Juan Jose

14imran:
We did it !

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,8 +20,60 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    let firstIteration = Recipe.create(data[0])
+     firstIteration
+        .then((res)=>{
+              console.log("data was sent",res)
+    })
+        .catch(()=>{
+              console.log("failure")
+    })
+  
+  
+  let secondIteration = Recipe.insertMany(data)
+    secondIteration
+        .then((res)=>{
+              res.filter((res)=>{ console.log(res.title)})
+      
+               console.log("data was sent")
+    })
+        .catch(()=>{
+              console.log("failure")
+      
+    })
+  
+
+
+  Promise.all([firstIteration,secondIteration])
+
+  .then(()=>{
+    let thirdIteration =Recipe.findOneAndUpdate({title:"Rigatoni alla Genovese"},{$set:{duration:100}})
+  thirdIteration
+          .then((res)=>{
+                console.log("data was sent",res)
+               })
+          .catch(()=>{
+                console.log("failure")
+               })
+    let fourthIteration =  Recipe.deleteOne({title:'Carrot Cake'})
+               fourthIteration
+                    .then((res)=>{
+                            console.log("data was sent",res)
+                         })
+                   .catch(()=>{
+                            console.log("failure")
+                         })
+                       })
+
+         Promise.all([thirdIteration, fourthIteration])
+          .then(() => {
+            mongoose.connection.close()
+            . then(() => {
+
+            })
+          })              
   })
   .catch(error => {
     console.error('Error connecting to the database', error);
   });
+  
--- a/models/Recipe.model.js
+++ b/models/Recipe.model.js
@@ -2,7 +2,38 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
-  // TODO: write the schema
+  title : {type:String,required:true},
+  level :{
+          type:String,
+          required:true , 
+          enum:['Easy Peasy',"Amateur Chef",'UltraPro Chef']
+        
+        },
+  ingredients : [String],//doubt
+  // ingredients: [
+  //   {
+  //     type: String
+  //   }
+  // ]
+  cuisine : String,
+  // cuisine: {
+  //   type: String
+  // }
+           
+  dishType : {
+    type:String,
+    enum:['breakfast','main_course','soup','snack', 'drink', 'dessert', 'other']
+  },
+  image: {type:String,
+  default: "https://images.media-allrecipes.com/images/75131.jpg"},
+  duration : {type:Number,
+              min:0},
+  creator : String ,
+  created :{type:Date,
+           default: Date.now },
+
+
+
 });
 
 const Recipe = mongoose.model('Recipe', recipeSchema);


models/Recipe.model.js
```diff
@@ -2,7 +2,38 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
-  // TODO: write the schema
+  title : {type:String,required:true},
+  level :{
+          type:String,
```
IH-WebDev-TA-Remote, line 7: check indentation here, you can also use the auto format feature of VScode with prettier to help with this

models/Recipe.model.js
```diff
@@ -2,7 +2,38 @@ const mongoose = require('mongoose');
...
+  cuisine : String,
+  // cuisine: {
+  //   type: String
+  // }
+           
+  dishType : {
+    type:String,
+    enum:['breakfast','main_course','soup','snack', 'drink', 'dessert', 'other']
+  },
+  image: {type:String,
```
IH-WebDev-TA-Remote, line 27: check code cleanness here, click enter after the "{" to start a new line for multiline code

models/Recipe.model.js
```diff
@@ -2,7 +2,38 @@ const mongoose = require('mongoose');
...
+  default: "https://images.media-allrecipes.com/images/75131.jpg"},
```
IH-WebDev-TA-Remote, line 28: and click enter before the "}" here to end the multiline code

models/Recipe.model.js
```diff
@@ -2,7 +2,38 @@ const mongoose = require('mongoose');
...
+  duration : {type:Number,
+              min:0},
+  creator : String ,
+  created :{type:Date,
```
IH-WebDev-TA-Remote, line 32: same as above

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
...
+    })
+  
+
+
+  Promise.all([firstIteration,secondIteration])
+
+  .then(()=>{
+    let thirdIteration =Recipe.findOneAndUpdate({title:"Rigatoni alla Genovese"},{$set:{duration:100}})
+  thirdIteration
+          .then((res)=>{
```
IH-WebDev-TA-Remote, line 52: since you are using promise.all later you don't need to check the promise resolution here as well with a .then and .catch. As this will already be handled inside the promise.all. It makes this code a little redundant.

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    let firstIteration = Recipe.create(data[0])
+     firstIteration
+        .then((res)=>{
```
IH-WebDev-TA-Remote, line 25: since you are using promise.all later you don't need to check the promise resolution here as well with a .then and .catch. As this will already be handled inside the promise.all. It makes this code a little redundant.

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
...
+              console.log("data was sent",res)
+    })
+        .catch(()=>{
+              console.log("failure")
+    })
+  
+  
+  let secondIteration = Recipe.insertMany(data)
+    secondIteration
+        .then((res)=>{
```
IH-WebDev-TA-Remote, line 35: same as above, this code is redundant as the .then and .catch are already being handled in the promise.all, so you can move this console.logs in there.

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
...
+              res.filter((res)=>{ console.log(res.title)})
```
IH-WebDev-TA-Remote, line 36: filter is used when you want to create a new array. If you just need to run a code for each element, then it is better to use the forEach method.

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    let firstIteration = Recipe.create(data[0])
```
IH-WebDev-TA-Remote, line 23: don't name your variables firstIteration, twoIteration and so on. It is better to put descriptive names for what will be stored inside the variable.

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
...
+              console.log("failure")
+      
+    })
+  
+
+
+  Promise.all([firstIteration,secondIteration])
+
+  .then(()=>{
+    let thirdIteration =Recipe.findOneAndUpdate({title:"Rigatoni alla Genovese"},{$set:{duration:100}})
```
IH-WebDev-TA-Remote, line 50: for mongoose methods that update the database you don't need to use $set. As the method always expect that second argument to be the one to be updated.

index.js
IH-WebDev-TA-Remote, line 52: same thing about code redundancy, these two promises are being checked inside the promise.all later. So using another .then and .catch here is not really needed. Instead you can put this part of the code (the console.log) inside the .then of the promise.all

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
...
+  thirdIteration
+          .then((res)=>{
+                console.log("data was sent",res)
+               })
+          .catch(()=>{
+                console.log("failure")
+               })
+    let fourthIteration =  Recipe.deleteOne({title:'Carrot Cake'})
+               fourthIteration
```
IH-WebDev-TA-Remote, line 59: check indentation, you can see that this variable is in another complete level of your code

index.js
```diff
@@ -20,8 +20,60 @@ mongoose
...
+                   .catch(()=>{
+                            console.log("failure")
+                         })
+                       })
+
+         Promise.all([thirdIteration, fourthIteration])
+          .then(() => {
+            mongoose.connection.close()
+            . then(() => {
+
```
IH-WebDev-TA-Remote, line 72: don't leave yout .then() empty. if you don't put anything in there then you don' need them in the first place. These .then and .catch are not needed for the promise to be executed. They only exist to run some code after the promise was either resolved or rejected.

IH-WebDev-TA-Remote:
The code works and is updating the database correctly, however there are several issues with code redundancy and code cleanness. 
Be more careful with indentation, and correct multiline code syntax as this will make your code easier to read not only to other developers but also to yourself when working on bigger projects. Check the auto-formatting feature of VScode with prettier that will help you a lot with that.

PR #1373, opened 2020-10-08
[REM-2009] Santiago & Laura

LauraSalakari:
c:

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,14 +29,18 @@ mongoose
     //     console.log("Something went wrong", err);
     //   });
 
-    let updateRecipe = Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese" }, { $set: { duration: 100 } });
-    let deleteRecipe = Recipe.deleteOne({ title: "Carrot Cake" });
+
     let insertRecipes = Recipe.insertMany(data);
-    let disconnectPromise = Promise.all([updateRecipe, deleteRecipe, insertRecipes]);
+    // let updateRecipe;
+    // let deleteRecipe;
 
     insertRecipes
       .then((res) => {
         console.log("Data added");
+
+        let updateRecipe = Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese" }, { $set: { duration: 100 } });
+        let deleteRecipe = Recipe.deleteOne({ title: "Carrot Cake" });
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -20,7 +20,70 @@ mongoose
...
+    insertRecipes
+      .then((res) => {
+        console.log("Data added");
+
+        let updateRecipe = Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese" }, { $set: { duration: 100 } });
+        let deleteRecipe = Recipe.deleteOne({ title: "Carrot Cake" });
+        let disconnectPromise = Promise.all([updateRecipe, deleteRecipe]);
+
+        updateRecipe
+          .then((res) => {
```
IH-WebDev-TA-Remote, line 46: the disconnectPromise holds the resolution of both the update and delete promises so running the console.logs separately here instead of inside the .then of disconnectPromise doesn't make much sense. To avoid code redundancy you can just declare and store the promises like you did in lines 41 and 42 and then put the code you want to run inside the .then and .catch of the promise.all

index.js
```diff
@@ -20,7 +20,70 @@ mongoose
...
+    //   });
+
+
+    let insertRecipes = Recipe.insertMany(data);
+    // let updateRecipe;
+    // let deleteRecipe;
+
+    insertRecipes
+      .then((res) => {
+        console.log("Data added");
```
IH-WebDev-TA-Remote, line 39: part of the iteration was to show the title of the elements inserted, using a loop on the res

index.js
```diff
@@ -20,7 +20,70 @@ mongoose
...
+            console.log("Recipe deleted", res);
+          })
+          .catch((err) => {
+            console.log("Could not delete", err);
+          });
+
+        disconnectPromise
+          .then((res) => {
+            console.log(res);
+            mongoose.disconnect();
```
IH-WebDev-TA-Remote, line 64: disconnect() is itself a promise, so the console.log below will run even if there was a problem disconnecting. To solve this, wrap it inside a .then attached to the disconnect()

IH-WebDev-TA-Remote:
Very good job! check specific code feedback for some issues with code redundancy. Also check the solution for promise.all we will send later.
But still overall, very well done. Keep it up! :)

PR #1420, opened 2020-11-05
[FTAMS102020] Rich Alossi

alossix:


LloydC:
Nice work Rich, I just wanted to mention that in your list of scripts for `package.json` you should set the value of 'dev' to 'nodemon index.js'.

Right now it's the same as your 'start' so it doesn't make sense to have it.

Everything else looks fine.

PR #1421, opened 2020-11-05
[FTAMS102020] Elia Chiarucci

eliachiarucci:


models/Recipe.model.js
```diff
@@ -2,7 +2,15 @@ const mongoose = require('mongoose');
...
-  // TODO: write the schema
+  title: {type: String, required: true, unique: true},
+  level: {type: String, enum: ["Easy Peasy", "Amateur Chef", "UltraPro Chef"]},
+  ingredients: {type: [String]},
+  cuisine: {type: String, required: true},
+  dishType: {type: String, enum: ["breakfast", "main_course", "soup", "snack", "drink", "dessert", "other"]},
+  image: {type: String, default: "https://images.media-allrecipes.com/images/75131.jpg"},
+  duration: {type: Number, min: 0},
+  creator: {type: String},
+  created: {type: Date, default: new Date()}
```
LloydC, line 13: You can also use `Date.now` instead of `new Date()`, you can see an example of it [here](https://mongoosejs.com/docs/schematypes.html#dates)

PR #1426, opened 2020-11-05
[FTAMS102020] Emma Northridge

emmanorthridge:
Completed lab - issues with mongo fixed

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,5 @@
 const mongoose = require('mongoose');
 
-// Import of the model Recipe from './models/Recipe.model.js'
 const Recipe = require('./models/Recipe.model');
 // Import of the data from './data.json'
 const data = require('./data');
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,27 +1,73 @@
...
+        {
+          new: true,
+        }
+      )
+        .then(() => console.log("Data is deleted"))
+        .catch((err) => console.error(`Error: ${err}`));
+
+      mongoose.connection.close();
+    });
   })
```
LloydC, line 51: Hi Emma, 

So the problem with your solution on line 21-49 is that you have several promises that are executing(create, insertMany, update and delete) at the same time and because each runs asynchronously there's a high chance that you won't be able "update" or "delete" because "insertMany" isn't completely done for example.

To prevent that a possible solution would be to have each promise happen in separates `.then()` so that each action would have to be completed before moving on to the next one.

Here's an example of what I mean
```
.then(() => {
    Recipe.create(recipeObj)
      .then((results) => console.log(`Saved new recipe: ${results}`))
      .catch((saveErr) => console.log(`Save failed: ${saveErr}`));
})
  .then(recipe => {
      .then((results) => console.log(`Saved new recipe: ${results}`))
      .catch((saveErr) => console.error(`Save failed: ${saveErr}`))
    console.log(recipe.title)
  })
  //ITERATION 4- Update recipe
  .then(recipe => {
    return Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese" }, { duration: 100 })
      .then(() => console.log(`Duration is updated`))
      .catch((saveErr) => console.error(`Save failed: ${saveErr}`));
  })
  .catch(error => {
  });

//ITERATION 5- Remove a recipe
  .then(recipe => {
    return Recipe.deleteOne({ title: "Carrot Cake" })
    .then(() => console.log(`Carrot Cake is deleted`))
    .catch((saveErr) => console.error(`Save failed: ${saveErr}`))
  })
//ITERATION 6- Close the Database
  .then(recipe => {
  return mongoose.connection.close()
      .then(() => console.log(`Database is closed!`))
      .catch(() => console.error('Delete Failed'))
    })
    .catch (error => {
      console.error('Error connecting to the database', error)
    })

emmanorthridge:
Thanks @LloydC .

I originally had multiple seperate .then() functions. 
However, I had some issues with the "findOneandUpdate" method, so Thomas advised me to put everything into one function.

I'll take a look again and see if I can work out the issue!

emmanorthridge:
@LloydC  Updated the code - hopefully should be correct now!

PR #1427, opened 2020-11-05
[FTAMS102020] Irene Asensio

ihatepineapple:


wax-bear:
Good morning Irene! :)
I see that quite some code is commented.
That are several ways to ensure the entirety of it is executed while taking into account the asynchronicity of the methods.
Please let me know if you would like to discuss those!

PR #1428, opened 2020-11-05
[WD-FT-AMS-102020] Yann Quintin

YannQuintin:


wax-bear (commented):
Good morning Yann! :) 
I see that quite some code is commented.
That are several ways to ensure the entirety of it is executed while taking into account the asynchronisity of the methods.
Please let me know if you would like to discuss those!

PR #1430, opened 2020-11-05
[FTAMS102020] AlejandroPliakasBenitez

apliakas:
some code is commented so it worked when I was testing it


wax-bear (commented):
Hey Alex, good stuff!
Would you mind submitting the iterations uncommented?
And is there perhaps a way to execute the methods after each other?

PR #1435, opened 2020-11-05
WDFT-AMS-102020]-Alex Kantsadis

teof95:
Done all 6 iterations


wax-bear (commented):
Great job Alex!
You chained it all with then methods, which is perfectly valid.
Please remove redundant comments if you can.

PR #1437, opened 2020-11-05
[AMSFT102020]issaGruber

GruberI:


index.js
```diff
@@ -20,8 +23,40 @@ mongoose
...
+        //iteration 5
+        Recipe.deleteOne({
+            title: 'Carrot Cake'
+          }, {
+            new: true
+          })
+          .then(() => console.log("data deleted"))
+          .catch(err => console.error(`Error: ${err}`));
+
+          mongoose.connection.close()
```
wax-bear, line 56: Due to this line, I do not see the results of update and delete in the db.
Could this method be placed somewhere where it waits for the completion of the .deleteOne method?

wax-bear (commented):
Well done Issa!
Please check out my comment regarding the connection.close method.

PR #1438, opened 2020-11-05
[FTAMS102020] Sabela Pena Garcia 

defSabe:
Completed it 1-2-3

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@ const mongoose = require('mongoose');
 
 // Import of the model Recipe from './models/Recipe.model.js'
 const Recipe = require('./models/Recipe.model');
-// Import of the data from './data.json'
+const recipesData = require("./data.json");
 const data = require('./data');
 
 const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -19,9 +19,171 @@ mongoose
...
+        image: "https://imagesvc.meredithcorp.io/v3/mm/image?url=https%3A%2F%2Fimages.media-allrecipes.com%2Fuserphotos%2F4398987.jpg&w=596&h=399.32000000000005&c=sc&poi=face&q=85",
+        duration: 30,
+        creator: "Chef Jennifer"
+      }
+    ];
+    Recipe.insertMany(arr, {
+        ordered: false
+      }),
+
+      //   function (error, docs) {});
```
wax-bear, line 145: Is this comment, among others, still useful?

index.js
```diff
@@ -19,9 +19,171 @@ mongoose
...
+
+
+      arr.findOneAndUpdate({
```
wax-bear, line 148: This method is not working as expected because it does not wait for the result of .inserMany. Either, place it in the body of the then method that is chained to the insertMany method, or use async/await in order to let .insertMany execute completely before .findOneAndUpdate is.

index.js
```diff
@@ -19,9 +19,171 @@ mongoose
...
+  //   }
+  //   return Recipe.create(recObj);
+  // })
+
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
-  })
-  .catch(error => {
-    console.error('Error connecting to the database', error);
+    const arr = [{
```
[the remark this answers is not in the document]
defSabe: not sure I understand... what would you suggest?

index.js
defSabe: true, probably they are not helpful anymore. Thanks for the feedback!

PR #1523, opened 2020-11-13
[REMOTE-PT-092020] Dirk Kelderman

dirkkelderman:


models/Recipe.model.js
```diff
@@ -2,7 +2,35 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
-  // TODO: write the schema
+  title: String,
```
In-S4n3, line 5: Here I would add an required tag, because every recipe must have a name right?!

title: { type: String, required: true}

models/Recipe.model.js
```diff
@@ -2,7 +2,35 @@ const mongoose = require('mongoose');
...
+  image: {
+    type: String,
+    default: "https://images.media-allrecipes.com/images/75131.jpg"
+  },
+  duration: {
+    type: Number,
+    min: 0
+  },
+  creator: String,
+  created: {
```
In-S4n3, line 30: noiiceeeee

index.js
```diff
@@ -20,7 +20,36 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+     return Recipe.create({ 
+      title: 'Chilli Con Carne',
```
In-S4n3, line 24: Love this recipe btw man! 👍

index.js
```diff
@@ -20,7 +20,36 @@ mongoose
...
+      level: 'Easy Peasy',
+      ingredients: 'minced beef, onions, garlic, tomatos, kidney beans',
```
In-S4n3, line 26: I would add rice also :D

index.js
```diff
@@ -20,7 +20,36 @@ mongoose
...
+  .then(() => {
+    return Recipe.deleteOne({title: 'Carrot Cake'})
+      // .then(recipe => console.log(`Recipe deleted`, recipe))
+      // .catch(error => console.log('An error happened while deleting the recipe:', error));
+  })
+  .then(() => {
+    mongoose.connection.close();
   })
   .catch(error => {
     console.error('Error connecting to the database', error);
```
In-S4n3, line 55: You made it like a pro super Dirk. Personaly, I would just not do so many then() methods and do more Recipe.create(), Recipe.insertMany() and so on.

Good job man, I better watch out! :D

PR #1531, opened 2020-11-16
done

ZuzanaAK:


models/Recipe.model.js
```diff
@@ -3,6 +3,38 @@ const Schema = mongoose.Schema;
...
+  duration: {
+    type: Number,
+    min: 0,
+  },
+  creator: String,
+  created: {
+    type: Date,
+    default: Date.now,
+  } 
+
```
In-S4n3, line 36: Everything here looks noiiiiiceeee :D

index.js
```diff
@@ -21,7 +41,62 @@ mongoose
   })
   .then(() => {
     // Run your code here, after you have insured that the connection was made
+      //for iteration 2
+    /*
```
In-S4n3, line 45: Not sure why everything is commented, bue everything looks good :). Good job!

ZuzanaAK: Because that's what they say in the instructions for iteration 2 and 3: "Tip: When you have successfully created a new recipe (you see it in the database using Compass tool), you might want to comment out this step. The reason for this is that next time when you run $ node index.js, it will try to create a new recipe with the same name and you will get an error in the terminal related to the duplicate keys - the title should be unique, and the dish with that title already exists in the database."

In-S4n3: Thats true, but you have commented all the iterations, or am I wrong?! Anyways! You did very well :)

ZuzanaAK: you are right, haha, I was experimenting with it to understand the concept. So, I was commenting out different pieces of the code, I guess I forgot to uncomment it when I was pushing it. Sorry about that!

PR #1536, opened 2020-11-17
Remote PT 09/2020

alesscamma:


models/Recipe.model.js
```diff
@@ -2,9 +2,34 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
-  // TODO: write the schema
+  title: String,
```
In-S4n3, line 5: The title should also be required, we would like to know what the name of that fantastic italian dish :D

index.js
```diff
@@ -20,8 +20,37 @@ mongoose
     return self.connection.dropDatabase();
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    return Recipe.insertMany(data)
```
In-S4n3, line 23: where's your recipe man?! I was hoping to learn some awesome past dish!

index.js
```diff
@@ -20,8 +20,37 @@ mongoose
...
+      })
+      .catch(error => {
+        console.log('An error occurred deleting the recipe', error);
+      });
+  })
+  .then(() => {
+    mongoose.connection.close();
+    console.log('Connection with the database closed');
+  })
+
```
In-S4n3, line 53: Overall looks very good! Personaly, I would not do so many then(). but more Recipe.create(), Recipe.insertMany() bla bla... it's more readable in my opinion.

Good job man!

PR #1720, opened 2021-04-15
[FTAMS032021] Silvia Rodriguez

SilviaRdgz:
On iteration 5,  I checked MongoDB compass and the recipe db was gone! Could we discuss how did this happen if I made no changes on the connection or deleted anything but one recipe ?

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,10 @@ mongoose
     .catch((err) => console.log(err));
 
 
-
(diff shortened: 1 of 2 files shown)


reMRKableDev (commented):
Good attempt @SilviaRdgz 🚀 

Let us have a chat this week about all the things that are at the moment a question mark or an uncertainty

PR #1722, opened 2021-04-15
[FTAMS032021] Andrea Sforza

arsforza:
Completed lab

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 const mongoose = require('mongoose');
+mongoose.set('useFindAndModify', false);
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,4 +1,5 @@
 const mongoose = require('mongoose');
+mongoose.set('useFindAndModify', false);
```
reMRKableDev, line 2: You can set this value in the `options` object in the `db.config.js` file. The same place where `useNewUrlParser` is passed

reMRKableDev (commented):
Good work @arsforza 🚀 

I left a comment you can check out

PR #1870, opened 2021-09-02
LAB_Mongoose_Recipes - Noemi Radvan WDPT JUL 2021

codingfannn:


IH-WebDev-TA-Remote:
Great job Noemi! 👏🏻👏🏻 Another lab done! Just a quick note, line 52 of index.js is returning you an `undefined`! The goal is to loop through every recipe, logging to the console only the title. Try to research what array methods you can use to achieve that! 💭

PR #1893, opened 2021-09-09
WD PT RM Chris Toth

Crixus90:


IH-WebDev-TA-Remote:
Almost there Chris! Regarding the log of each recipe's title, you can use a forEach loop on the manyRecipes, logging the title of each one inside the loop. Then (hint!), you can continue with iteration 4 & 5. Good luck! 🍀

PR #1917, opened 2021-10-14
[FT-RMT-092021] Dominika Lipka

dominikalipka:
At first it was challenging but then I understood the pattern and it was easier to follow. I know that probably I shouldn't console.log in a separate ".then" but my brain wasn't working no more this evening 😪


luisjunco:
Bonus:
- use Promise.all() (the operations to update and delete do not need to happen "one after the other"; instead, you can request both operations and close the connection as soon as both are completed).
- translate all the code to async/await.

PR #1920, opened 2021-10-15
[FT-RMT-092021] Alice

AliceXix:
Had a little trouble with accessing the variable data in .then

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,6 +37,12 @@ mongoose
   .then( () => {
     return console.log(`Youhuouuuu this worked!`)
   })
+  .then( () => {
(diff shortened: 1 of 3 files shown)


TA-Remote:
Great work on the lab! The only note I would give, is that you do not need a separate ` .then() `  for your 
`console.log`

eg. 

`.then(newRecipe => {
    console.log(`we have recieved a new recipe ${newRecipe.title}`)
  })
  .then((allRecipes) => {
    console.log(`We have now ${allRecipes} in the database`)
    return Recipe.findOneAndUpdate({ title: 'Rigatoni alla Genovese' }, { $set: { duration: 100}}, {new: true})
  })`

PR #1969, opened 2021-11-12
[FTLIS102021] Anne 

annemedan:
all changes have been made - async await file also created


pmcgarrett:
Great Job, using the then() and the async await function!

PR #2146, opened 2022-04-21
WDFT REMOTE MAR22 Mafalda & Fevi

fevi-ro:


TA-Remote:
good job! For the date, try to use Date.now() to set the current date instead of having to hardcode it yourselves.
Also, try and use catch statements always to protect our code.
Finally, when creating the recipe, we could try and add more items instead of only three

PR #2147, opened 2022-04-21
[WDFTMAR2022] Ivo Silva - Oscar Gonzalez

Ivo-S-Silva:
Completed all 6 iterations.


TA-Remote:
Great job as always! Next time try and use Date.now() when setting a date so you don't have to hardcode it! :D

PR #2150, opened 2022-04-21
WDFT MAR22 Harry & Daniel

hmarchbank:


TA-Remote:
Great job guys, just one tiny comment: when setting a default date, you could call the Date.now method and get the current date instead of having to hardcode it yourself

PR #2152, opened 2022-04-21
WDFTMarch2020 Valentin Paul & Pedro Palma

Valentin-Paul:
solved iterations


TA-Remote:
good job! Try to console.log the title after creating the recipe, and use the forEach in iteration 3! :D

PR #2210, opened 2022-06-03
almost solved

EvgeniiaRiaguzova:


Enes-Ironhack:
Iteration 1 => OK
2 => Commented so not working
3 => You were supposed to use the Data variable that imports the data from the json file.
4 => Ok
5 => OK
6 => Commented so KO

PR #2213, opened 2022-06-04
not finished

nataliabuzila:


Enes-Ironhack:
Iteration 1 => OK
2 => OK
3 => Trying to add a recipe that already exists.  and you were supposed to use the DATA variable.
4 => OK
5 => Ok
6 => KO

PR #2242, opened 2022-07-07
[RM FT June 22] Alex's Lab amost done, one iteration missing

Virgule09:
One iteration missing and the bonus but I need to run to the doctors for my back :).
I hope that I can finish it later but I am not sure :/

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,6 +32,18 @@ mongoose
     const update = { duration: 100 };
     return Recipe.findOneAndUpdate(filter, update);
   })
+  .then( () => {
+    console.log("Succes the recipe has been upadted !")
+  }) 
+  .then( () => {
+    return Recipe.deleteOne({title: 'Carrot Cake' })
+  }) 
+  .then( () => {
+    console.log("Succes the recipe was deleted")
+  }) 
+  .then( () => {
+    mongoose.connection.close()
+  })
   .catch(error => {
     console.error('Error connecting to the database', error);
   });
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -11,13 +11,40 @@ const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
...
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    // iteration 2
+    // Recipe.create(data[0]);
+
+    return Recipe.insertMany(data);
+  })
+  .then( result => {
+    return Recipe.find({duration :{$gte:0}} ,"title" );
+  })
```
TA-Remote, line 26: here you dont need to use find() method,,< you can get the result (this is an array) and you can iterate this array with forEach or every loop you want and print every title of element of result array :
` .then(() => {
  })
  .then((recipesArr) => {
    recipesArr.forEach((recipe) => {
      console.log(recipe.title);
    })`

index.js
```diff
@@ -11,13 +11,40 @@ const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
...
+  .then(titles => {
+    console.log(titles)
+  } )
+  .then(() => {
+    const filter = { title: "Rigatoni alla Genovese" };
+    const update = { duration: 100 };
+    return Recipe.findOneAndUpdate(filter, update);
+  })
+  .then( () => {
```
TA-Remote, line 35: best practice is returning a promise here ! << to keep promise chain ;)
you can write
` return Recipe.deleteOne({title: 'Carrot Cake' })`
inside this .then ()

PR #2243, opened 2022-07-07
Lara Rockenstein // first iterations done together with Christoph // WD FT June22

LMRBLN:


TA-Remote (commented):
Great work!
Just couple of notes 

It seems you have skipped an iteration..

For iteration 2 - you had to create a new recipe object based on the schema you wrote, and send it to the db

For iteration 3 - using Recipe.insertMany, send all the data to the db

PR #2245, opened 2022-07-07
Khalid Alsehsah & Paul Tiry [WDFT JUN2022]

ptiry02:
happy cooking! 😉


TA-Remote:
Great work!
Just one little note, for iteartion2, you should have created your own recipe object (invent a recipe, rather than grab one from data ), based on the schema you defined

PR #2247, opened 2022-07-07
Begum Eryildiz WDFTJune2022

begumeryildiz:
Solved the first four iterations with Alex, then I solved the last two iterartions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,34 @@ mongoose
   })
   .then(() => {
     // Run your code here, after you have insured that the connection was made
+
+    return Recipe.create(data[0]);
+
+  })
+  .then(createdRecipe => {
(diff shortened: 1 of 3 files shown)


TA-Remote:
Great Job Begum!

index.js
```diff
@@ -17,6 +17,34 @@ mongoose
   })
   .then(() => {
     // Run your code here, after you have insured that the connection was made
+
+    return Recipe.create(data[0]);
```
TA-Remote, line 21: here you created a recipe with first element of data, but in the Recipe model,, title should be unique. it means in line 27 when you wanna create your database with data file,, you have a duplicate title,, and you get an error,,<

PR #2248, opened 2022-07-07
Tobias Conrad [WDJUN22FT]

maktooob:


index.js
```diff
@@ -17,6 +17,27 @@ mongoose
...
+   return Recipe.insertMany(data) // Iteration 3
+  })
+  .then((result) => {
+    result.forEach(element => console.log(element.title)) // Iteration 3
+    return Recipe.findOneAndUpdate({
+    title: "Rigatoni alla Genovese"
+    },
+    {
+      duration: 100
+    })
```
TA-Remote, line 30: Here you could added the `{ new: true }` and you would get the updated document

models/Recipe.model.js
```diff
@@ -3,6 +3,24 @@ const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
   // TODO: write the schema
+  title : String,
+  level: String,
```
TA-Remote, line 7: On instructions the model had more info like:

```
title: {
  type: String,
  required: true,
  unique: true
}
```

TA-Remote (commented):
- Great work

PR #2251, opened 2022-07-07
WD RT FT JUN 2022 | Ruben Andino

rubenanlo:
Completed assessment. Thanks in advance for your review!!

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ const mongoose = require('mongoose');
 const Recipe = require('./models/Recipe.model');
 // Import of the data from './data.json'
 const data = require('./data');
+const { db } = require('./models/Recipe.model');
 
 const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -16,8 +17,133 @@ mongoose
...
   })
   .then(() => {
-    // Run your code here, after you have insured that the connection was made
+    return Recipe.create({title: 'margharita', level: 'difficult', ingredients: ['tomato', 'mozzarella', 'ham', 'cheese'], cuisine: 'Italian', dishType: 'pizza', duration: 30, creator: 'Ruben'})
+  })
+  .then(() => {
+
+    app.get('./data', (req, res, next) => { ; //--> to retrieve data. The code included in here is a middleware.
+      res.send("prodcuts");
+    })
```
TA-Remote, line 26: That is throwing an error, the `app` in other examples is a calling from `express` that hasn't been defined or installed here and there is a `;` that would cause problems too => All the data is already imported at `const data = require('./data');`

index.js
```diff
@@ -16,8 +17,133 @@ mongoose
...
+  //   "dishType": "dessert",
+  //   "image": "https://imagesvc.meredithcorp.io/v3/mm/image?url=https%3A%2F%2Fimages.media-allrecipes.com%2Fuserphotos%2F4398987.jpg&w=596&h=399.32000000000005&c=sc&poi=face&q=85",
+  //   "duration": 30,
+  //   "creator": "Chef Jennifer"
+  // }
+  // ]
+    return Recipe.insertMany(data);
+  })
+  .then(() => {
+    console.log(Recipe.find({projection: {title: 1}}))
```
TA-Remote, line 126: - Here you're trying to `console.log()` a promise, I think you wanted to show the results of that promise, if that's the case you needed to `return` the promise and `console.log()` in a `.then()` after that one. 

- In the `Model.find()` the first argument is the criteria that you want to use as the search and the second one the projection => `Recipe.find( { },   {title: 1} )` I think that would `return` what you were expecting => Or at least would search for all recipes and show just the `titles` and `_id` :)

models/Recipe.model.js
```diff
@@ -2,6 +2,21 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const recipeSchema = new Schema({
+  title: String,
```
TA-Remote, line 5: The fields for the Model should have more infos as described on instructions

PR #2254, opened 2022-07-08
Jack Le

jackletran:
finished all iterations, not sure if 6 is correct though. Please correct me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,13 +40,26 @@ mongoose
   })
   .then((recipes) => {
     // console.log(recipes);
-    Recipe.findOneAndUpdate(
+    console.log("The faulty recipe was successfully changed!");
+
+    return Recipe.findOneAndUpdate(
       {
         title: "Rigatoni alla Genovese",
       },
-      { duration: 100 }
+      { duration: 100 },
+      { returnDocument: "after" }
     );
-    console.log("The faulty recipe was successfully changed!");
+  })
+  .then((recipeToRemove) => {
+    console.log("A recipe was removed successfully!");
+
+    return Recipe.deleteOne({ title: "Carrot Cake" });
+  })
+  .then((result) => {
+    console.log("The database was closed.");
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,23 +1,66 @@
...
+  })
+  .then((recipeToRemove) => {
+    console.log("A recipe was removed successfully!");
+
+    return Recipe.deleteOne({ title: "Carrot Cake" });
+  })
+  .then((result) => {
+    console.log("The database was closed.");
+    return mongoose.end(MONGODB_URI);
+
```
TA-Remote, line 61: mongoose.disconnect()

index.js
```diff
@@ -1,23 +1,66 @@
...
+    // console.log("Recipe was created!");
+
+    // return Recipe.create(newProduct);
+
+    console.log("Multiple recipes were created!");
+
+    return Recipe.insertMany(data);
+  })
+  .then((recipes) => {
```
TA-Remote, line 41: In the `.then()` you are always passing an argument, but you're never using them, so you can remove them from your code if you want and it could be better for readability :)

index.js
```diff
@@ -1,23 +1,66 @@
...
+    //   ingredients: "MSG, Thai Curry Paste, Peanut Butter, Coconut Milk",
+    //   cuisine: "Asian",
+    //   dishType: "main_course",
+    //   duration: 60,
+    //   creator: "Jack Le Chef",
+    // };
+
+    // console.log("Recipe was created!");
+
+    // return Recipe.create(newProduct);
```
TA-Remote, line 35: Instead of creating a newProduct you could have used the `data[0]` for example

TA-Remote (commented):
- Great Work!!

- Pay attention to bits of code you aren't using
- Your console logs in general are appearing before the actions they announce will be executed, that could cause a false impression that the action already took place

PR #2255, opened 2022-07-08
WD RMT June 2022 Matthieu Duperrey

MatthieuDUPERREY:


index.js
```diff
@@ -17,6 +17,26 @@ mongoose
...
+  })
+  .then((result) => {
+    result.forEach(element => console.log(element.title)) // Iteration 3
+    return Recipe.findOneAndUpdate({
+      title: "Rigatoni alla Genovese"
+    },
+      {
+        duration: 100
+      })
+  })
```
TA-Remote, line 30: Here you could have added the `options object` (the 3rd argument for that method) at the end to see the result after updating the documento on the DB

PR #2256, opened 2022-07-08
up to iteration 4 -darren

darrenmd:


index.js
```diff
@@ -13,12 +13,136 @@ mongoose
   .then(x => {
     console.log(`Connected to the database: "${x.connection.name}"`);
     // Before adding any recipes to the database, let's remove all existing ones
-    return Recipe.deleteMany()
+    return Recipe.deleteMany();
   })
-  .then(() => {
+  console.log(Recipe.title)
+  .then(() =>  {
```
TA-Remote, line 19: You cannot have any code in between `.then()`

think of it as connected chain:

`mongoose.connect().then().then().then() `etc

But now you added console.log which will break your code

`mongoose.connect().then()consolelog().then().then() `

index.js
```diff
@@ -13,12 +13,136 @@ mongoose
...
+    //     "duration": 30,
+    //     "creator": "Chef Jennifer"
+   //   });
+    return Recipe.insertMany([
+      {
+        "title": "Asian Glazed Chicken Thighs",
+        "level": "Amateur Chef",
+        "ingredients": [
+          "1/2 cup rice vinegar",
+          "5 tablespoons honey",
```
TA-Remote, line 45: At the top of this file we are importing all the data, so you can simply pass that array directly

`const data = require('./data');`

`Recipe.insertMany(data)`

PR #2340, opened 2022-09-08
Solved Lab

avarap:


index.js
```diff
@@ -1,23 +1,66 @@
...
     // Before adding any recipes to the database, let's remove all existing ones
-    return Recipe.deleteMany()
+    return Recipe.deleteMany();
   })
-  .then(() => {
+  .catch((error) => {
+    console.error("Error connecting to the database", error);
+  });
+
+mongoose
```
trebeljahr, line 47: this is somewhat broken. 
you are calling mongoose.connect twice - and you don't properly wait for the first connection to be done deleting all there is in the database. this could lead to possible raceconditions. instead either -> use async and await or do whatever you need to do within chained .then blocks after the return Recipe.deleteMany()

index.js
```diff
@@ -1,23 +1,66 @@
...
+  .connect(MONGODB_URI)
+  .then(async (x) => {
+    console.log(`Connected to the database: "${x.connection.name}"`);
     // Run your code here, after you have insured that the connection was made
+    await recipe1.save();
```
trebeljahr, line 52: if you use async await -> don't mix them inside of a .then block please. this is inconsistent and somewhat ugly code. sure it works, but why not make everything async / await or everything promise based?

trebeljahr: i.e. be consistent with yourself, if possible.

models/Recipe.model.js
```diff
@@ -2,7 +2,49 @@ const mongoose = require('mongoose');
...
+    type: Number,
+    default: 0,
+  },
+  creator: {
+    type: String,
+  },
+  created: {
+    type: Date,
+    default: Date.now,
+  },
```
trebeljahr, line 47: the Schema looks good :)

trebeljahr (commented):
good job. :) 

only complaint is that you are mixing and matching things (promises/async await) somewhat inconsistently. if you mix and match, at least stay consistent with your own code.

PR #2343, opened 2022-09-09
[PTBER072022] berndHavekost

hvkst:
that went much better than expected :)


indexES.mjs
```diff
@@ -0,0 +1,51 @@
...
+import data from './data.json' assert { type: 'json' };
+const MONGODB_URI = 'mongodb://localhost:27017/recipe-app';
+
+// await mongoose.connect('mongodb://localhost:27017/pokemons-v3');
+// await mongoose.connection.db.dropDatabase();
+// await mongoose.connection.close();
+
+// await mongoose.connect('mongodb://localhost:27017/pokemons-v3');
+
+async function main() {
```
trebeljahr, line 20: no need for a async main function in a .mjs file -> top level await is a thing here :)

models/Recipe.model.js
```diff
@@ -2,7 +2,41 @@ const mongoose = require('mongoose');
...
+  },
+  cuisine: {
+    type: String,
+    required: true,
+  },
+  dishType: {
+    type: String,
+    enum: ['breakfast', 'main_course', 'soup', 'snack', 'drink', 'dessert', 'other'],
+  },
+  hp: {
```
trebeljahr, line 25: should this really be named hp?

trebeljahr (commented):
really good work, as usual. :+1: keep it up. and get better soon :tada:

PR #2378, opened 2022-11-02
done

cursedxp:


cursedxp:
I added close connection in one of my commits

PR #2534, opened 2023-03-16
lab done with some issues on iteration 4 but the code seems to be fine

andrehcarvalho6:


TA-Remote:
Good job🙌!!!
maybe it was a bit confusing, because you've never seen how to 'required files'. 
Just want to make sure you got it now🙂
`
// Import of the data from './data.json'
const data = require("./data");  // --> we require the file here, and we store it's content in the variable 'data'

// your code
...
  .then(() => {
    const recipesArr = require("./data.json"); // we've already imported at the top😉
    Recipe.insertMany(recipesArr); // add the return so that at the next .then you can check if the 'data' array has been added
    recipesArr.forEach((recipe) => { // this console.log display the title of the array, not the result from DB
      console.log(recipe.title);
    });
    return Recipe.findOneAndUpdate( // in this method you can pass field and value you are looking for {title: 'Rigatoni...'}
      "6413590efe79f46683f1cedf",
      // {title: "Rigatoni alla Genovese"},

      {
        duration: 100,
      },
      { new: true }
    );
  })
...
// possible solution
  })
  .then(() => {
    return Recipe.insertMany( data )
    })
  .then((recipesAddedFromDB) => {
     recipesAddedFromDB.forEach((recipe) => {
           console.log(recipe.title)
           })
    return Recipe.findOneAndUpdate({title: 'Rigatoni alla Genovese'}, {duration: 100}, {new:true})
    })
    .then((updatedRecipe) => {
      console.log(`the new duration for ${updatedRecipe.title} is ${updatedRecipe.duration}`)
      return ......
    })
`
hope it'll be helpful, in any case you know where to find me🫶

PR #2646, opened 2023-07-06
[PT] Xavier & Maria 

MariaG6:


index.js
```diff
@@ -1,23 +1,63 @@
...
+    });
   })
-  .catch(error => {
-    console.error('Error connecting to the database', error);
+  .then((newRecipe) => {
+    console.log(newRecipe.title);
+    return Recipe.insertMany(data);
+  })
+  .then((manyRecipes) => {
+    console.log(manyRecipes);
```
Web-Dev-PT-EN, line 44: You don't need a `.then()` just for the `console.log()` so you could have returned the next promise directly here
You could also use the `.forEach()` to print just the titles :) 
```suggestion
    manyRecipes.forEach(recipe => console.log(recipe.title))
      return Recipe.findOneAndUpdate(
      { title: "Rigatoni alla Genovese" },
      {duration:100},
      { new: true }
    );
```

Web-Dev-PT-EN (commented):
Good work, keep practicing :) 

If you want more practice with promises you can convert the lab to `async/await`

PR #2647, opened 2023-07-07
Thor - Agustin

tngknzr:


index.js
```diff
@@ -17,7 +17,57 @@ mongoose
...
+  .then(updatedRecipe => {
+    return Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese" }, { duration: 100 })
+
+  })
+  .then(() => {
+    console.log('Updated succesfull')
+  })
+
+  .then((outcake) => {
+    Recipe.deleteOne({ title: "Carrot Cake" })
```
Web-Dev-PT-EN, line 54: Remember to return the operation with the database so you can use in the next `.then()`

index.js
```diff
@@ -17,7 +17,57 @@ mongoose
...
+  })
+  .then(() => {
+    console.log(`titles, ${data[0].title}`);
+  })
+  .then(updatedRecipe => {
+    return Recipe.findOneAndUpdate({ title: "Rigatoni alla Genovese" }, { duration: 100 })
+
+  })
+  .then(() => {
+    console.log('Updated succesfull')
```
Web-Dev-PT-EN, line 50: You don't need a `.then()` for just a `console.log()`
```suggestion
    console.log('Updated succesfull')
    return Recipe.deleteOne({ title: "Carrot Cake" })
```

